Issue #1, opened 2024-07-18
Set up CI


Tom-Willemsen:
As a developer I would like a CI setup in github actions for this repository:

### On PRs and new commits in `main`:
- Linting 
  * Should include at least as much as in https://github.com/ISISComputingGroup/reusable-workflows , but as this is a brand-new repository we may prefer to be stricter around docstrings and types in particular
  * There may also be additional `ruff` checks - particularly around generators / async - that will help us for bluesky-specific code.
- Unit tests - pytest
- As this repo may be used by other groups in future (e.g. mantid), run the above checks in GH actions on most recent 3 python versions (following SPEC00 recommendation)
- Block PRs being merged if above tests and linting don't pass
  * As this is a brand new repo we should not need to just check diffs, the whole repo should pass all of our checks.

### On new release
- Linting as above
- Tests as above
- If linting and tests pass, publish release to PyPI
  * Will need to set up an ISISComputingGroup PyPI  org, using shared mailbox
  * Ensure any keys/tokens/passwords/secrets get added to keeper
- Use standard github "create release" infrastructure for this.
- Create a `0.0.1` or similar release to prove this all works and document release process in README/wiki/etc

# Acceptance criteria:
- [x] Depends on https://github.com/ISISComputingGroup/IBEX/issues/8439, which must have been completed
- [x] PRs checked by linting and tests
- [x] Releases checked by linting and tests
- [x] After a successful release, the new code is available to any user via pip (using `python -m pip install ibex_bluesky_core` or another suitable name).

## Time in planning meeting (18/07/24)
1:15:20

LowriJenkins:
https://github.com/ISISComputingGroup/ibex_bluesky_core/pull/4
To review, use pip to install ibex_bluesky_core to a venv, then follow the instructions in `demo_plan.py` to run the demo plan

PR #4, opened 2024-07-23
Setup ci

LowriJenkins:
https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/1


.github/workflows/Linter.yml
```diff
@@ -0,0 +1,38 @@
+name: Linter
+on: [pull_request]
+jobs:
+  call-workflow:
+    strategy:
+      matrix:
+        version: ['3.10','3.11','3.12']
```
Tom-Willemsen, line 7: If lint & test are separate tasks (which it looks like they will be here) then I think we don't need to run linters on all 3 python versions, can pick just one.

Sorry, this is my fault I wasn't super precise on the ticket, I was imagining "lint & test" being one job, and that job would then need to be run on all 3 pythons, but what you have in this PR is probably a better approach.

jackbdoughty (dismissed):
All looks good to me but couple things that I noticed:
- Not sure if its been changed since but the Keeper username to PyPi isn't the same as the email address so needs updating
- Noticed that in the acceptance criteria that that 'As this repo may be used by other groups in future (e.g. mantid), run the above checks in GH actions on most recent 3 python versions (following SPEC00 recommendation)' from what I understand this supports 3.10 and 3.11 not 3.12 but this is waiting on another ticket to be merged before possible. Worth keeping an eye on this
- Regarding the readme instructing how to create a release, it took me a couple extra steps as I had to find where/how to get the release approved in GH so maybe worth making it slightly clearer in the instructions :)

jackbdoughty (approved):
Changes look good.

Issue #2, opened 2024-07-18
Create a run engine


Tom-Willemsen:
We will need a function to acquire a suitably-configured bluesky run engine under which plans will be executed

- This should be a lazily instantiated singleton (only one run engine should ever be created in any given python process).
  * The run engine should _not_ be created at library import time.
  * Implementation hint: `@functools.cache`
- This method should be part of the "public" API of this library
  * In a future ticket, we may add something like `RE = ibex_bluesky_core.create_run_engine()` as one of the GUI initialization commands which are automatically run. However that is out-of-scope for _this_ issue.
- Consider how the bluesky event loop interacts with the existing matplotlib event loop in the GUI
  * Generally multiple event loops are not usually recommended, read up on best practice in this area to decide what to do in each of these cases.
  * e.g. make sure you read https://github.com/bluesky/ophyd-async/blob/ef1c9426cb20c597a7614686aa01cbf9d43d11ac/docs/explanations/event-loop-choice.rst and other similar material
- We already use an `asyncio` event loop for matplotlib plotting (in `genie_python.matplotlib_backend`):
  * Review that event loop setup/configuration to make sure the two event loops don't clash with each other / check any global configuration around event loop policies and `asyncio` error handlers.
- For _this_ issue, we do not need to add any particular hooks or customization to the run engine, these will be added in a future ticket. 
- Ensure basic run engine control works - especially keyboard interrupts, you should be able to pause/resume/abort a long-running plan (e.g. `bps.sleep(99)`) both from a normal python shell and from the IBEX gui.

# Acceptance criteria

Something in the following form works:
```python
import bluesky.plan_stubs as bps
from bluesky.run_engine import RunEngine
from ibex_bluesky_core import get_run_engine

RE: RunEngine = get_run_engine()
RE(bps.null())  # Or any other plan - null() is the simplest possible plan, which emits no messages

ANOTHER_RE: RunEngine = get_run_engine()  # Should return the same run_engine as before
assert ANOTHER_RE is RE
```

Ensure unit tests have been created, to cover at a minimum:
- Ensuring we don't create multiple run engines
- Basic run engine functionality - running a trivial plan
- Ability to pause/resume/abort the run engine (this is standard bluesky functionality)
- Ability to get the plan result out of the run engine (this is standard bluesky functionality)

## Time in planning meeting (18/07/24)
1:26:00 (ish - sorry, GR)

Tom-Willemsen:
PRs:
[...]

PR #5, opened 2024-07-25
Create run engine

Tom-Willemsen:
Create a run engine, and add basic tests for run engine functionality.

https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/2

To review:
- Check demo plan works
- Check tests & linting pass (this will soon be done by CI, but as CI is not merged yet, it needs to be checked manually)
- Check as a developer you agree with the initial basic set of tests for bluesky run engine functionality (the goal is not to exhaustively test it, but to check we've configured it in a sane way)
- Check as a developer basic interactive run engine functionality works - e.g. run a plan, interrupt it, add a custom callback, etc.

Notes:
- The custom `_DuringTask` has been pushed upstream (https://github.com/bluesky/bluesky/pull/1770); if & when that PR gets merged & released we can remove our version

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,3 +79,5 @@ directory = "coverage_html_report"
 
 [tool.pyright]
 include = ["src", "tests"]
+reportUntypedFunctionDecorator = true
+reportUntypedClassDecorator = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ dev = [
 [tool.pytest.ini_options]
 testpaths = "tests"
 asyncio_mode = "auto"
-addopts = "--cov --cov-report=html"
+addopts = "--cov --cov-report=html -vv"
 
 [tool.coverage.run]
 branch = true
--- a/src/ibex_bluesky_core/demo_plan.py
+++ b/src/ibex_bluesky_core/demo_plan.py
@@ -5,13 +5,15 @@ from typing import Generator
 import bluesky.plan_stubs as bps
 from bluesky.callbacks import LiveTable
 from bluesky.preprocessors import run_decorator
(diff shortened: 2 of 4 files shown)


iangillingham-stfc:
Everything this looks pretty good and all tests look valid and pass. I just have a couple of points/questions that cropped up in my review:
1) I question the instance naming of the RunEngine class as "RE". It flies in the face of PEP-8 and is also the same name as the 're' module (regular expressions), apart from the case.
2) I struggled to see how a single instance of the RunEngine was guaranteed to be singleton - I'm assuming it's the @cache decorator that's producing this effect? Did you consider creating a true singleton subclass of RunEngine?

Otherwise, the code is really tidy and well documented.

iangillingham-stfc (dismissed):
Very tidy and well documented. I left some comments in the PR, but nothing that should prevent this being merged.

Tom-Willemsen:
As discussed in person:

> I question the instance naming of the RunEngine class as "RE". It flies in the face of PEP-8 and is also the same name as the 're' module (regular expressions), apart from the case.

The thought was to follow the bluesky conventions here, e.g. the [docs, tutorial](https://nsls-ii.github.io/bluesky/tutorial.html) and most other bluesky facilities use `RE` for their runengines.

> I struggled to see how a single instance of the RunEngine was guaranteed to be singleton - I'm assuming it's the `@cache` decorator that's producing this effect? Did you consider creating a true singleton subclass of RunEngine?

Comment added to further clarify the existing docstring on the function.

iangillingham-stfc (dismissed):
The documented explanation of the use of @cache has clarified the get_run_engine() function returning a singleton.

iangillingham-stfc (approved):
Looks good. All tests in place and pass.

Issue #6, opened 2024-07-30
Callbacks: document logging


Tom-Willemsen:
As a developer, it would be useful to capture the "raw" documents coming out of the run engine, so that an exact stream of documents can be "replayed" into future callbacks to test them.

This also ensures that all information emitted by bluesky is captured.

### Acceptance criteria
- Callback added to default run engine which saves data for each bluesky "run"
  * **NOTE: "bluesky run" != "ibex run", a run in bluesky encompasses a whole scan**
- Likely logging into `c:\instrument\var\logs\bluesky\raw_documents` or a similar location
- It should be possible to retrieve the **full** content of every bluesky document after-the-fact, by inspecting these log files.
  * With the idea that we can later "replay" these same documents into other callbacks to test them.
- This callback should write it's output incrementally as it captures documents, so that if a bluesky plan is hard-aborted halfway through, there is still a way to retrieve documents that were emitted before the abort.
- This callback is non-optional, added by default in `get_run_engine()` for example.

### Notes
- Longer term we will probably do this via e.g. kafka instead so these simple files will likely become redundant.
- Add directory to log rotation script (move rather than delete)
- Bluesky has facilities for this (e.g. `databroker`) that log into a database, could consider using those facilities, but equally consider doing something very simple to start with. The primary use case for this callback will be developer debugging so consider which workflow gives the best developer experience.

### Time in Planning
00:28:28 08/08/24

jackbdoughty:
[EPICS PR](https://github.com/ISISComputingGroup/EPICS/pull/361)
[BlueSky core PR](https://github.com/ISISComputingGroup/ibex_bluesky_core/pull/11)

PR #11, opened 2024-08-09
Added document logging & unit tests

jackbdoughty:
- Creates a new callback function that the BlueSky run engine subscribes to which records every document in a log file under "C:\\","instrument","var","logs","bluesky","raw_documents".
- Added a unit test that checks that the file to be written to is opened 3 times, for open, descriptor then stop. Also in the stop document to be written, checks that the run is successful with no interruptions.
- Callback returns an ID of 0 if it is ever needed to unsubscribe

[Ticket6](https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/6#issue-2437829032)

### DIFF
--- a/src/ibex_bluesky_core/callbacks/document_logger.py
+++ b/src/ibex_bluesky_core/callbacks/document_logger.py
@@ -8,33 +8,28 @@ log_location = Path(os.path.join("C:\\", "instrument", "var", "logs", "bluesky",
 
 
 class DocLoggingCallback:
-    """
-    Logs all documents under log_location, with the file name of their UID (.log).
-    """
+    """Logs all documents under log_location, with the file name of their UID (.log)."""
 
     def __init__(self) -> None:
-        """
-        Initialises current_start_document and filename.
-        """
-
+        """Initialise current_start_document and filename."""
         self.current_start_document = None
         self.filename = None
 
     def __call__(self, name: str, document: dict) -> int:
-        """
-        Is called when a new document needs to be processed. Writes document to a file.
+        """Is called when a new document needs to be processed. Writes document to a file.
 
         Args:
             name: The type of the document e.g start, event, stop
             document: The contents of the docuement as a dictionary
+        
         """
-
         if name == "start":
             log_location.mkdir(parents=True, exist_ok=True)
 
             self.current_start_document = document["uid"]
-            self.filename: str = os.path.join(log_location, f"{self.current_start_document}.log")
+            self.filename = os.path.join(log_location, f"{self.current_start_document}.log")
 
+        assert self.filename is not None, "Could not create filename."
         assert self.current_start_document is not None, "Saw a non-start document before a start."
 
         to_write = {"type": name, "document": document}
--- a/src/ibex_bluesky_core/callbacks/document_logger.py
+++ b/src/ibex_bluesky_core/callbacks/document_logger.py
@@ -21,7 +21,6 @@ class DocLoggingCallback:
         Args:
             name: The type of the document e.g start, event, stop
             document: The contents of the docuement as a dictionary
-        
         """
         if name == "start":
             log_location.mkdir(parents=True, exist_ok=True)
--- a/src/ibex_bluesky_core/callbacks/document_logger.py
+++ b/src/ibex_bluesky_core/callbacks/document_logger.py
@@ -21,6 +21,7 @@ class DocLoggingCallback:
         Args:
             name: The type of the document e.g start, event, stop
             document: The contents of the docuement as a dictionary
+
         """
         if name == "start":
             log_location.mkdir(parents=True, exist_ok=True)
--- a/src/ibex_bluesky_core/callbacks/document_logger.py
+++ b/src/ibex_bluesky_core/callbacks/document_logger.py
@@ -1,0 +1,34 @@
+"""Logs all documents that the BlueSky run engine creates via a callback"""
+
+import os, json
+from pathlib import Path
+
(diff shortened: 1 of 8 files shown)


src/ibex_bluesky_core/callbacks/document_logger.py
```diff
@@ -0,0 +1,40 @@
+"""Logs all documents that the BlueSky run engine creates via a callback."""
+
+import json
+import os
+from pathlib import Path
+
+log_location = Path(os.path.join("C:\\", "instrument", "var", "logs", "bluesky", "raw_documents"))
```
Tom-Willemsen, line 7: Since we're using `pathlib` anyway...

```suggestion
log_location = Path("C:\\") / "instrument" / "var" / "logs" / "bluesky" / "raw_documents"
```

src/ibex_bluesky_core/callbacks/document_logger.py
```diff
@@ -0,0 +1,40 @@
...
+        Args:
+            name: The type of the document e.g start, event, stop
+            document: The contents of the docuement as a dictionary
+
+        """
+        if name == "start":
+            log_location.mkdir(parents=True, exist_ok=True)
+
+            self.current_start_document = document["uid"]
+            self.filename = os.path.join(log_location, f"{self.current_start_document}.log")
```
Tom-Willemsen, line 30:
```suggestion
            self.filename = log_location / f"{self.current_start_document}.log"
```

src/ibex_bluesky_core/callbacks/document_logger.py
```diff
@@ -0,0 +1,40 @@
...
+
+class DocLoggingCallback:
+    """Logs all documents under log_location, with the file name of their UID (.log)."""
+
+    def __init__(self) -> None:
+        """Initialise current_start_document and filename."""
+        self.current_start_document = None
+        self.filename = None
+
+    def __call__(self, name: str, document: dict) -> int:
```
Tom-Willemsen, line 18: Does this actually need to return an `int`? I didn't think callbacks returned anything but maybe I'm wrong...

src/ibex_bluesky_core/run_engine.py
```diff
@@ -65,4 +67,7 @@ def get_run_engine() -> RunEngine:
     )
     RE.record_interruptions = True
 
+    log_callback = DocLoggingCallback()
+    RE.subscribe(log_callback)  # Uses ID = 0 for DocLoggingCallback
```
Tom-Willemsen, line 71: I thought it was the RE that would tell you what your callback ID is (which might change if in future we add some other callback before this one...)

Not that we actually _care_ for this - I think the ID is only needed if you want to unsubscribe a callback, which we don't want to for this one.

tests/test_run_engine.py
```diff
@@ -139,3 +142,30 @@ def test_during_task_does_wait_with_small_timeout():
 
     event.wait.assert_called_with(0.1)
     assert event.wait.call_count == 2
+
+
+def test_run_engine_logs_all_documents(RE):
+    m = mock_open()
+    filepath: str
+    log_location = Path(
+        os.path.join("C:\\", "instrument", "var", "logs", "bluesky", "raw_documents")
```
Tom-Willemsen, line 151: As per other suggestions: use pathlib fully.

tests/test_run_engine.py
```diff
@@ -139,3 +142,30 @@ def test_during_task_does_wait_with_small_timeout():
...
+    def basic_plan() -> Generator[Msg, None, None]:
+        yield from bps.open_run()
+        yield from bps.close_run()
+
+    with patch("ibex_bluesky_core.callbacks.document_logger.open", m):
+        result: RunEngineResult = RE(basic_plan())
+        filepath = os.path.join(log_location, f"{result.run_start_uids[0]}.log")
+
+    for i in range(0, 3):
+        assert m.call_args_list[i].args == (filepath, "a")
```
Tom-Willemsen, line 163: There's probably a nicer way to do this using `assert_calls` or something similar in the `unittest.mock` library?

tests/test_run_engine.py
```diff
@@ -139,3 +142,30 @@ def test_during_task_does_wait_with_small_timeout():
 
     event.wait.assert_called_with(0.1)
     assert event.wait.call_count == 2
+
+
+def test_run_engine_logs_all_documents(RE):
```
Tom-Willemsen, line 147: I wonder whether this test would be better in a new `test_document_logging_callback.py` file?

The usual idea is to have a roughly 1-1 correspondence between implementation files and test files.

(we can chat about this if you want - it'll involve moving the `RE` fixture into a `conftest.py`)

src/ibex_bluesky_core/callbacks/document_logger.py
```diff
@@ -0,0 +1,40 @@
+"""Logs all documents that the BlueSky run engine creates via a callback."""
```
Tom-Willemsen, line 1: I think the new `callbacks` module should have an `__init__.py` to be importable

Tom-Willemsen (commented):
Could you write some docs (just some markdown in the `docs` folder) about what this does, how it works, where the logs go, how to use it for debugging etc?

tests/test_run_engine.py
jackbdoughty: Tried to do this but got difficult

Tom-Willemsen (approved):
Looks good, thanks.

We might need to do something cleverer if we ever end up doing nested multi-run plans, but this is fine for the simple plans we'll be doing in the immediate future.

PR #27, opened 2024-09-03
9 add basic dae signals

### LINKED ISSUE #9: DAE: add basic signals
As a user I would like access to various typically-used configuration signals on the DAE.

Most signals which are settable via the "DAE configuration" screen in IBEX, or readable as "interesting" PVs from the DAE, should be in scope.

For this issue, the signals need to be added to the DAE ophyd-async device, as either "standard" or "configuration" signals as appropriate for each signal.

--- 

In scope:
- Most readable "interesting" float/int/str/enum PVs served directly from the DAE
  * Ensure this includes metadata like RB number, users, run number, ...
- Most writeable configuration float/int/str/enum PVs served by the DAE.
- Spectra, but limit to a single, constant "interesting" spectrum at a time for now
  * Prove we _can_ read a spectra data array, don't worry about details of how it will be used in practice in this issue.

Out of scope (for this issue):
- Derived signals (e.g. normalising intensity by monitor counts)
- Filtered signals (e.g. regions-of-interest on spectra)
- Anything that goes via areadetector (e.g. live view - this to be the subject of future issue(s))
- Higher-level logic, e.g. the exact sequence of bluesky messages to begin/acquire/end a data point at the "top level".

### Time in Planning
00:49:50 08/08/24

rerpha:
closes #9 

This PR adds basic DAE signals. the `Dae` device is intended to be subclassed in the future as it currently does nothing when triggering. 
All basic readbacks from the GUI should be available, with the exception of spectra - currently we only show spectra 1 period 1, this is because we don't want to subscribe to everything by default.
All settings in the GUI should also be available too - these are stored in XML on the ICP end, so there are some utilities to deal with this when reading/setting to convert them to a nicer format.


pyproject.toml
```diff
@@ -42,8 +42,9 @@ classifiers = [
 
 dependencies = [
   "bluesky",
-  "ophyd-async[ca]",
+  "ophyd-async[ca]@git+https://github.com/bluesky/ophyd-async.git",
```
Tom-Willemsen, line 45: Can we add a comment explaining which PR(s) we're waiting for to be released?

src/ibex_bluesky_core/demo_plan.py
```diff
@@ -36,14 +36,15 @@ def demo_plan() -> Generator[Msg, None, None]:
     @run_decorator(md={})
     def _inner() -> Generator[Msg, None, None]:
         # A "simple" acquisition using trigger_and_read.
-        yield from bps.abs_set(block, 1.0, wait=True)
-        yield from bps.trigger_and_read([block, dae])
+        # Currently this does not work as the DAE does not have a trigger method.
+        # yield from bps.abs_set(block, 1.0, wait=True)
+        # yield from bps.trigger_and_read([block, dae])
```
Tom-Willemsen, line 41: I think just remove if it flat out doesn't work.

src/ibex_bluesky_core/devices/dae/__init__.py
```diff
@@ -0,0 +1,51 @@
+"""Utilities for the DAE device - mostly XML helpers."""
+
+from __future__ import annotations
```
Tom-Willemsen, line 3: Did we actually need this?

https://peps.python.org/pep-0649/ is accepted and deprecates https://peps.python.org/pep-0563/ ...

src/ibex_bluesky_core/devices/dae/dae_controls.py
```diff
@@ -0,0 +1,46 @@
...
+
+        super().__init__(name=name)
+
+
+class BeginRunExBits(IntFlag):
+    """Bits for BEGINRUNEX."""
+
+    NONE = 0
+    BEGINIFPAUSED = 1
+    BEGINIFDELAYED = 2
```
Tom-Willemsen, line 32: I think this is begin _in_ paused mode not begin _if_ paused mode? Double check...

src/ibex_bluesky_core/devices/dae/dae_period_settings.py
```diff
@@ -0,0 +1,132 @@
...
+    frames: int | None = None
+    output: int | None = None
+    label: str | None = None
+
+
+@dataclass
+class DaePeriodSettingsData:
+    """Dataclass for the hardware period settings."""
+
+    periods_settings: List[SinglePeriodSettings]
```
Tom-Willemsen, line 58: I guess this means we _always_ have to specify periods_settings - is there a way around that? It feels like we should be able to accept `None` for this?

src/ibex_bluesky_core/devices/dae/dae_settings.py
```diff
@@ -0,0 +1,172 @@
...
+VETO0_NAME = "Veto 0 Name"
+MUON_CERENKOV_PULSE = "Muon Cerenkov Pulse"
+MUON_MS_MODE = "Muon MS Mode"
+FC_WIDTH = "FC Width"
+FC_DELAY = "FC Delay"
+VETO3 = "Veto 3"
+VETO2 = "Veto 2"
+VETO1 = "Veto 1"
+VETO0 = "Veto 0"
+# Yes, there is a space prefixing some of the vetos
```
Tom-Willemsen, line 32: Wow that's incredibly horrible.

src/ibex_bluesky_core/devices/dae/dae_tcb_settings.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class TimeRegimeMode(Enum):
+    """Time Regime Mode options for a single row."""
+
+    BLANK = 0
+    DT = 1
+    DTDIVT = 2
+    DTDIVT2 = 3
+    SHIFTED = 4
```
Tom-Willemsen, line 49: These could probably benefit from a small comment, I have no idea what `DTDIVT2` means off the top of my head...

tests/devices/test_dae.py
```diff
@@ -26,36 +59,2461 @@ def test_explicit_dae_naming():
...
-    assert "DAE-good_uah" in reading.keys()
+    assert reading.keys() == descriptor.keys()
 
 
-async def test_dae_describe_contains_intensity_and_default_keys(dae: Dae):
-    descriptor = await dae.describe()
+async def test_begin_run_sets_begin_run_ex(dae: Dae):
+    get_mock_put(dae.controls.begin_run_ex.begin_run_ex).assert_not_called()
+    await dae.controls.begin_run_ex.set(BeginRunExBits.NONE)
+    get_mock_put(dae.controls.begin_run_ex.begin_run_ex).assert_called_once()
```
Tom-Willemsen, line 76: This should probably check it was called with value 0?

Tom-Willemsen (commented):
Just a few quick comments after a skim-read...

src/ibex_bluesky_core/devices/dae/__init__.py
rerpha: don't think so, IDE probably added it

src/ibex_bluesky_core/devices/dae/dae_period_settings.py
rerpha: yes, https://github.com/ISISComputingGroup/ibex_bluesky_core/pull/27/commits/22fc9fc8c7babe2d1a8e983a2217bac8d1097893

Tom-Willemsen (approved):

PR #34, opened 2024-10-03
Fitting Callback

jackbdoughty:
Introduces our own data fitting callback as a wrapper to Bluesky's LiveFit. 
- Adds extra functionality such as guessing heuristics, each extra piece of data will mean new guess generated before trying to fit. Plus new fitting models over what already was in scans library.
- Unit tests for each type of fit and the callback itself.
- Documentation.

Best way to test would be to go through all new docs. Then to spin up some fake noisy data and set/read values to blocks. 
```
while True:
    val = g.get_pv("CS:SB:bob", is_local=True)
    g.set_pv("CS:SB:alice",  lorentzian(val, 1, 1, 0),is_local=True)
```

Then create a plan in a separate file and check that fits are working. 
```
alice = block_r(float, "alice")
    bob = block_rw(float, "bob", write_config=BlockWriteConfig(settle_time_s=0.5))

    lf = LiveFit(Lorentzian.fit(), "alice", "bob", )

    fig, ax = plt.subplots()
    
    @subs_decorator(
        [
            LiveTable([bob.name, alice.name]),
            LiveFitPlot(lf, ax=ax, color="r", num_points=1000),
            LivePlot(y=alice.name, x=bob.name, ax=ax, marker="x", linestyle="none"),
        ]
    )
        yield from ensure_connected(bob, alice)
        yield from bp.scan([alice], bob, -5, 5, num=15)
    yield from _inner()
```

- Also worth checking that unit tests are appropriate and cover what they need to.
- And can check that fit results can be read from LiveFit.result


doc/fitting/fitting.md
```diff
@@ -0,0 +1,201 @@
...
+from ibex_bluesky_core.callbacks.plotting import LiveFit
+```
+**Note:** that you do not *need* `LivePlot` for `LiveFit` to work but it may be useful to know visaully how well the model fits to the raw data.
+
+## Configuration
+
+Below is a full example showing how to use standard `matplotlib` & `bluesky` functionality to apply fitting to a scan, using LivePlot and LiveFit. The fitting callback is set to expect data to take the form of a gaussian.
+```py
+import matplotlib.pyplot as plt
+from ibex_bluesky_core.callbacks.plotting import LivePlot, LiveFit
```
LowriJenkins, line 17: Correct imports

doc/fitting/fitting.md
```diff
@@ -0,0 +1,201 @@
...
+# Create a new figure to plot onto.
+plt.figure()
+# Make a new set of axes on that figure
+ax = plt.gca() 
+# ax is shared by fit_callback and plot_callback 
+
+plot_callback = LivePlot(y="y_variable", x="x_variable", ax=ax)
+fit_callback = LiveFit(Gaussian.fit(), y="y_variable", x="x_variable", update_every=0.5)
+# update_every = in seconds, how often to recompute the fit. If `None`, do not compute until the end. Default is 1.
+fit_plot_callback = LiveFitPlot(fit_callback, ax=ax, color="r")
```
LowriJenkins, line 28: Show the import of LiveFitPlot with the others to make it clear that this is from bluesky not ibex bluesky

LowriJenkins (changes requested):
Just some minor documentation changes

LowriJenkins (approved):

PR #36, opened 2024-10-10
21 generate dae uncertainties

jackbdoughty:
Exposes signals for summed dae counts uncertainties.

To test;
- Check that all unit tests pass.
- Check that the `dae_test` manual test runs and that the uncertainty is the sqrt of normalised counts.


src/ibex_bluesky_core/devices/simpledae/reducers.py
```diff
@@ -74,12 +79,16 @@ async def reduce_data(self, dae: "SimpleDae") -> None:
...
         self._det_counts_setter(float(summed_counts.value))
-        self._intensity_setter(float(summed_counts.value) / denominator)
 
-        detector_counts_var = 0.0 if summed_counts.variance is None else summed_counts.variance
+        if denominator == 0.0:
+            self._intensity_setter(0.0)
+            intensity_var = 0.0
+        else:
+            self._intensity_setter(float(summed_counts.value) / denominator)
+            temp_intensity_var = (summed_counts / denominator).variance
```
Tom-Willemsen, line 88: We're doing `summed_counts / denominator` twice here.

I think we should only do it once, and then extract `.value` and `.variance` from it at the end.

rerpha (approved):

rerpha:
@Tom-Willemsen  could you approve pls, ive shot myself in the foot by committing and now can't merge

Tom-Willemsen (approved):

PR #37, opened 2024-10-10
Add callback for human-readable-file writing

### LINKED ISSUE #7: Callbacks: file writing (human-readable files)
As a user I would like the results of my scan to be saved as a text file.

To do this, we need a callback which can be added to the runengine configured to write files.

### Acceptance criteria
- Callback added to default run engine which saves data for each bluesky run
  * **NOTE: "bluesky run" != "ibex run"**, a run in bluesky encompasses a whole scan which may be any number, including zero, if IBEX runs.
- Header of the file should contain:
  * Bluesky scan identifier
  * Time of scan start (from timestamp of the `open_run` document)
  * RB number - likely should be added by a metadata decorator on the run.
  * Metadata passed into the RE e.g. `RE(plan(), sample_id='A', purpose='calibration', operator='Dan', some_random_metadata_key='blah')`
  * Metadata passed explicitly into plan e.g. `yield from count([det], md={'purpose': 'calibration'})`
  * Metadata added by metadata wrappers
  * Metadata passed in by any other [standard bluesky md mechanisms](https://nsls-ii.github.io/bluesky/metadata.html)
- Body of the file should be in a textual n-column format
  * Should be possible to explicitly configure fields to write, but by default this should be the _hinted_ fields from the scanned variable(s) and detector(s).
- For this issue, do not worry about logging fit types or fit results
  * Fitting will be the subject of future issues.
- It should be possible for users to configure the directory into which these files are written
- See existing `LiveTable` callback, which is quite similar but prints data to screen instead of to file, for inspiration.
- For this issue, do not worry too much about _exact_ file formats - we will likely need to add further options to control exact file formats in future.

### Example file

Simplified file for this issue (don't worry about matching this format exactly, as long as the key information is there and the file remains human-readable).

This is designed to be close to a format that has already been discussed with scientists (see scans requirements doc for details).

```
scan_id: ffff-ffff-ffff-ffff
start_time: 2025-01-01 01:00:00 UTC
rb_number: 00000000
some_metadata: foo
some_more_metadata: bar

run_number    motor_position    dae_goodframes
00000001      42                420
00000002      73                0
00000003      123               123
```

We wll likely end up writing the data in a proper machine-readable format (e.g. `hdf5`) in future as well as these simple text files - these should prioritise human-readability over machine readability. (this is out of scope for _this_ issue)

### Time in Planning
00:34:23 08/08/24

rerpha:
Closes #7 

Adds a callback to make human-readable files with metadata and provided fields. More info in the added docs.


Tom-Willemsen:
Showed this to scientists, they have changed their mind and would now prefer the main table to be `csv` rather than `tsv`. Hopefully this is just a 1-character change...

doc/callbacks/file_writing.md
```diff
@@ -0,0 +1,6 @@
+# File writing callbacks
```
Tom-Willemsen, line 1: I feel this documentation could be expanded a little bit.

e.g. give a full example of how to use this callback including setting the directory to put the files into and controlling which columns to write out?

doc/callbacks/file_writing.md
```diff
@@ -0,0 +1,6 @@
...
+## Human readable files
+
+A callback (`HumanReadableOutputFileLoggingCallback`) exists to write all documents to a separate human-readable file which contains the specified fields.
```
Tom-Willemsen, line 4: nit: wrong name now

src/ibex_bluesky_core/callbacks/file_logger.py
```diff
@@ -0,0 +1,108 @@
...
+
+        This involves creating the file if it doesn't already exist
+        then putting the metadata ie. start time, uid in the header.
+        """
+        self.output_dir.mkdir(parents=True, exist_ok=True)
+        self.current_start_document = doc[UID]
+        self.filename = self.output_dir / f"{self.current_start_document}.txt"
+
+        exclude_list = [
+            TIME,  # We format this later
```
Tom-Willemsen, line 53: 👍

tests/callbacks/test_write_log_callback.py
```diff
@@ -0,0 +1,159 @@
...
+    time = 1728049423.5860472
+    uid = "test123"
+    scan_id = 1234
+    run_start = RunStart(time=time, uid=uid, scan_id=scan_id)
+    with patch("ibex_bluesky_core.callbacks.file_logger.open", mock_open()) as mock_file:
+        cb.start(run_start)
+        result = save_path / f"{run_start['uid']}.txt"
+
+    mock_file.assert_called_with(result, "a")
+    # assert not mock_file().write.assert_any_call(f"scan_id: {scan_id}")
```
Tom-Willemsen, line 30: Do we need this anymore?

tests/callbacks/test_write_log_callback.py
```diff
@@ -0,0 +1,159 @@
...
+    cb.filename = Path("test")
+
+    # This just contains the value
+    event = Event(uid=uid, data={field_name: expected_value}, descriptor=descriptor, seq_num=1)
+    with patch("ibex_bluesky_core.callbacks.file_logger.open", mock_open()) as mock_file:
+        cb.event(event)
+
+    mock_file.assert_called_with(cb.filename, "a", newline="")
+    first_call = call(f"\n{field_name}({units})\n")
+    second_call = call(f"{expected_value:.{prec}f}\r\n")
```
Tom-Willemsen, line 74: Hmm this looks like it's terminating some lines with `\n` and some with `\r\n`? I looked in a real output file and that also seems to be true there, that doesn't seem right.

We should at least be consistent about which line ending we're using.

tests/callbacks/test_write_log_callback.py
```diff
@@ -0,0 +1,159 @@
...
+    assert cb.descriptors[desc["uid"]] == desc
+
+
+def test_event_prints_header_with_units_and_respects_precision_of_value_on_first_message():
+    field_name = "test"
+    cb = HumanReadableFileCallback(save_path, [field_name])
+    # This actually contains the precision
+    expected_value = 1.2345
+    units = "mm"
+    prec = 4
```
Tom-Willemsen, line 60: Since `expected_value` is only to 4 d.p. we're not actually testing it gets formatted to 4 d.p. as far as I can tell?

tests/conftest.py
```diff
@@ -3,8 +3,15 @@
...
 from ibex_bluesky_core.run_engine import get_run_engine
 
+MOCK_PREFIX = "UNITTEST:MOCK:"
+
 
 @pytest.fixture
 def RE() -> RunEngine:
     get_run_engine.cache_clear()
-    return get_run_engine()
+    re = get_run_engine()
```
Tom-Willemsen, line 12: purely convention, and I know it flies in the face of PEP8, but bluesky uses `RE` for the runengine everywhere rather than `re`...

Tom-Willemsen (commented):
Looking pretty much there, small handful of mostly nit-picky comments and then I'll shut up I promise :)

Tom-Willemsen (approved):

Issue #23, opened 2024-08-22
Configure logging


Tom-Willemsen:
Read:
- https://blueskyproject.io/bluesky/main/debugging.html
- https://docs.python.org/3/howto/logging.html#

And then:
- Configure python loggers for `ibex_bluesky_core` following best practices in the documents linked above
- Add some initial debug logging in "obvious places" e.g. in DAE and blocks devices, ISIS-specific callbacks, describing what they're doing.

We need to do this so that we can get appropriate debugging information to support this system when it's in use on the instruments.

## Acceptance criteria:
- [ ] `ibex_bluesky_core` can use python's stdlib `logging` module to log stuff
- [ ] The log messages from both `bluesky` itself and `ibex_bluesky_core`, at a suitable log level, end up in `c:\instrument\var\logs\bluesky\`
- [ ] Ensure log file gets rotated every day so that log rotation scripts can move it
- [ ] Documentation about how to increase the log level(s) if we need more information to diagnose a problem
- [ ] Above location added to log rotation script

Chsudeepta:
The following PRs have been created:
Changes for logger: https://github.com/ISISComputingGroup/ibex_bluesky_core/pull/40
Changes for log archiving script: https://github.com/ISISComputingGroup/EPICS/pull/371
Changes for system tests: https://github.com/ISISComputingGroup/system_tests/pull/114

One caveat is that many system tests were failing, including the new ones added here. But these tests ran fine from in isolation (executed as unit tests).

Tom-Willemsen:
Can we write a follow-up issue for the second acceptance criteria:
> Add some initial debug logging in "obvious places" e.g. in DAE and blocks devices, ISIS-specific callbacks, describing what they're doing.

Since it doesn't seem to be in the PRs above?

PR #40, opened 2024-10-16
Ticket 23 configure bluesky logging

Chsudeepta:
Added blue sky logger, unit tests, logging configuration and documentation.
This is for the [ticket](https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/23)


doc/logging/logger.md
```diff
@@ -0,0 +1,7 @@
+# Bluesky logger
+
+To invoke the bluesky logger, import as `from ibex_bluesky_core.logger import logger` and use it at the desired level:
+`logger.blueskylogger.warning("Message to be logged")`
+The logger utilizes a `TimedRotatingFileHandler` defined in the `logging.conf` file that rolls over the log at midnight.
+
+The default logging level is defined at `INFO`.
```
jackbdoughty, line 7: Nitpicking a bit here sorry- I think we should provide some more info on how to change the log level as this is one of the acceptance criteria. e.g Where do I find info on what different log levels there are? And where do I make this change?

jackbdoughty (changes requested):
Other than the comment I left regarding docs looks good to me

doc/logging/logger.md
Chsudeepta: The logging levels are sort of python standard - DEBUG, INFO, WARNING etc. And we invoke them as --> mylogger.warning("message"), mylogger.error("another message"). I think it might be overkill to describe the python standard logging levels and calls here. 
As described here, the log handler is defined in logging.conf and that's where the default logging level is set to INFO for the handler

jackbdoughty: Okay cheers thats all fine then, merging now

jackbdoughty (approved):

PR #44, opened 2024-10-28
Fixes issues found with linear fitting guess function

jackbdoughty:
Found issues with linear fitting whereby the guess function would not work as expected e.g not guessing a straight line and division by zero error. This fixes by instead using the already implemented polynomial guess function and passing 1 as the polynomial degree. Should test by creating an example plan and two blocks to scan against eachother from- use a python script to repeatedly write to one of the blocks to simulate data from scan. See below. Check that initial guess produced in plot window fits the data provided well.

See [fitting docs](https://isiscomputinggroup.github.io/ibex_bluesky_core/fitting/fitting.html) for more info.

Run this separately

```
def linear(x, m, c):
    return m * x + c

g.set_pv("CS:SB:bob", 1, is_local=True)

while True:

    val = g.get_pv("CS:SB:bob", is_local=True)
    g.set_pv("CS:SB:alice", linear(val, 0, 1),is_local=True)
```

Use a plan that looks like this

```
def plan():

    alice = block_r(float, "alice")
    bob = block_rw(float, "bob", write_config=BlockWriteConfig(settle_time_s=0.5))

    lf = LiveFit(Linear.fit(), "alice", "bob")

    fig, ax = plt.subplots()
    
    @subs_decorator(
        [
            LiveTable([bob.name, alice.name]),
            LiveFitPlot(lf, ax=ax, color="r", num_points=1000),
            LivePlot(y=alice.name, x=bob.name, ax=ax, marker="x", linestyle="none"),
        ]
    )
        yield from ensure_connected(bob, alice)
        yield from bp.scan([alice], bob, -5, 5, num=15)
        logger.blueskylogger.info("Some useful message")
    yield from _inner()
```

### DIFF
--- a/src/ibex_bluesky_core/callbacks/fitting/fitting_utils.py
+++ b/src/ibex_bluesky_core/callbacks/fitting/fitting_utils.py
@@ -179,8 +179,8 @@ class Linear(Fit):
     def model(cls, *args: int) -> lmfit.Model:
         """Linear Model."""
 
-        def model(x: npt.NDArray[np.float64], m: float, c: float) -> npt.NDArray[np.float64]:
-            return m * x + c
+        def model(x: npt.NDArray[np.float64], c1: float, c0: float) -> npt.NDArray[np.float64]:
+            return c1 * x + c0
 
         return lmfit.Model(model)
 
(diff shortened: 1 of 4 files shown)


tests/callbacks/fitting/test_fitting_methods.py
```diff
@@ -219,7 +219,7 @@ def test_linear_model(self):
             grad = 3
             y_intercept = 2
 
-            outp = Linear.model().func(x, m=grad, c=y_intercept)
```
Tom-Willemsen, line 222: can we add a unit test for the case which was previously failing that you've now fixed?

src/ibex_bluesky_core/callbacks/fitting/fitting_utils.py
```diff
@@ -179,8 +179,8 @@ class Linear(Fit):
     def model(cls, *args: int) -> lmfit.Model:
         """Linear Model."""
 
-        def model(x: npt.NDArray[np.float64], m: float, c: float) -> npt.NDArray[np.float64]:
-            return m * x + c
+        def model(x: npt.NDArray[np.float64], c1: float, c0: float) -> npt.NDArray[np.float64]:
```
Tom-Willemsen, line 182: I think the docs now need updating to match these updated parameter names?

Tom-Willemsen: I also noticed the docs use `a` `b` `c` `d` for polynominal parameters but the code uses `c0`, `c1`, ...

Issue #43, opened 2024-10-27
Document how to call external code from within a plan


Tom-Willemsen:
As a user I would like to know how to call external code from within a plan, and what limitations are imposed when I do that.

---

It is important that simply iterating a plan does not do any i/o, so for example this would be incorrect:

```python
def bad_plan():
    yield from bps.open_run(...)
    ...
    inst.my_clever_inst_script(1, 2, 3)  # Bad - must not do this!
    ...
    yield from bps.close_run(...)
```

Instead, we need to trigger the relevant code using a bluesky `Msg`, and teach the `RunEngine` how to deal with it. We also need to make sure it's wrapped up in a coroutine if it's not already.

```python
async def _my_clever_inst_script_handler(msg: Msg):
    # Usual caveats of asyncio.to_thread apply: won't be cancellable, ...
    return await asyncio.to_thread(inst.my_clever_inst_script, *msg.args, **msg.kwargs)

RE.register_command("my_clever_inst_script", _my_clever_inst_script_handler)

def better_plan():
    yield from bps.open_run(...)
    ...
    yield from bps.clear_checkpoint()  # Also need to document when clear_checkpoint is and isn't needed
    yield Msg("my_clever_inst_script", 1, 2, 3)
    ...
    yield from bps.close_run(...)


# Note: above code is untested but should give the general idea...
```

### Notes
- We should _heavily discourage_ doing this where we've provided functionality natively in bluesky, with explanations that doing this loses:
  * Rewindability (which breaks e.g. ability to ctrl-c and then later choose to continue the scan seamlessly, as bluesky no longer "understands" what has been done)
  * Error handling in general is worse with this approach compared to native bluesky, e.g. task cancellation won't work properly. For this reason, we should limit these to routines that run "quickly". This approach is **definitely unsuitable** for anything which may get stuck waiting and which the user may want to ctrl-c.
  * Ability to integrate with any bluesky signals / emit documents / generally use most bluesky functionality
  * Any possible concurrent execution of this and other work in parallel
  * Despite the above caveats, I anticipate this being needed for e.g. `g.send_sms()` (which isn't too bad if it can't be interrupted, as it should be fast) and _maybe_ things like `do_sans`/`to_trans` on sans instruments.
[...]

PR #47, opened 2024-11-05
Add helper for external blocking functions

Tom-Willemsen:
https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/43

Contrary to what the ticket said, I did actually end up writing some code rather than just documentation here. My initial plan was to simply not allow keyboard interrupts, but on further thought I think a best-efforts mechanism, while not _perfect_ is better than nothing.

There's some fiddly code in this PR so please review carefully and make sure it is understandable for the average developer.

### DIFF
--- a/doc/plan_stubs/external_code.md
+++ b/doc/plan_stubs/external_code.md
@@ -1,5 +1,7 @@
 # `call_sync` (calling external code)
 
+API reference: {py:obj}`ibex_bluesky_core.plan_stubs.call_sync`
+
 All interaction with the "outside world" should be via bluesky messages, and **not** directly called from
 within a plan. For example, the following is **bad**:
 
--- a/doc/plan_stubs/external_code.md
+++ b/doc/plan_stubs/external_code.md
@@ -1,0 +1,75 @@
+# `call_sync` (calling external code)
+
+All interaction with the "outside world" should be via bluesky messages, and **not** directly called from
+within a plan. For example, the following is **bad**:
+
+```python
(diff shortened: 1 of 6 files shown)


src/ibex_bluesky_core/plan_stubs/__init__.py
```diff
@@ -1 +1,42 @@
...
+T = TypeVar("T")
+
+
+CALL_SYNC_MSG_KEY = "ibex_bluesky_core_call_sync"
+
+
+def call_sync(func: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> Generator[Msg, None, T]:
+    """Call a synchronous user function in a plan, and returns the result of that call.
+
+    Attempts to guard against the most common pitfalls of naive implementations, for example:
```
rerpha, line 18:
```suggestion
    Attempts to guard against the most common pitfalls of naive implementations, for example:
    
```
needed by MyST to render the bullet points

doc/plan_stubs/external_code.md
```diff
@@ -0,0 +1,75 @@
+# `call_sync` (calling external code)
+
```
rerpha, line 2: I think we should link to the API ref here. 

```suggestion

API reference: {py:obj}`ibex_bluesky_core.plan_stubs.call_sync`

```

as per https://myst-parser.readthedocs.io/en/latest/syntax/code_and_apis.html#sphinx-autodoc2

rerpha (dismissed):

PR #52, opened 2024-11-07
Use uncertainties in plots and fits

### LINKED ISSUE #22: Use uncertainties in plots and fits
As a user, if uncertainties have been published by the underlying device, I would like plots & fits to use them.

This issue depends on all of:
- https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/21
- https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/19
- https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/8

Which must be completed first.

### Acceptance criteria:
- [x] Uncertainties, if available, are plotted as error bars on plots.
  * We probably want a `LivePlotWithUncertainty` wrapper on top of `LivePlot` for this.
  * `matplotlib` can already handle plots with error bars, so this is largely "pass data through to matplotlib in the right places".
- [x] Uncertainties, if available, are used in fits. 
  * `lmfit` takes `weights` for each point which can be used to implement this - so this is largely "pass data through to lmfit in the right places".

rerpha:
Closes #22

### DIFF
--- a/doc/architectural_decisions/005-variance-addition.md
+++ b/doc/architectural_decisions/005-variance-addition.md
@@ -1,0 +1,17 @@
+# Why do we add 0.5 to variance?
+
+## Status
+
+Current
+
+## Context
+
+If we pass counts data to `scipp` as variance this is correct- but if one of the counts is 0 then its variance is 0, leading to a division by 0 error when calculating its weight for fitting `weight = 1 / doc["data"][self.yerr]`.
+
+## Decision
+
+Our solution was to add 0.5 (`VARIANCE_ADDITION`) to each count to calculate variance. The actual data should be unchanged, the +0.5 is only for uncertainty calculation.
+
+## Justification
+
+The above approach is both "smooth" and converges towards sqrt(N) in the limit with high counts, and should also mean that we never get an uncertainty of zero in the fitting side.
--- a/doc/architectural_decisions/005-variance-addition.md
+++ b/doc/architectural_decisions/005-variance-addition.md
@@ -6,7 +6,7 @@ Current
 
 ## Context
 
-If we pass counts data to `scipp` as variance this is correct- but if one of the counts is 0 then its variance is 0, leading to a division by 0 error when calculating its weight for fitting `weight = 1 / doc["data"][self.yerr]`.
+If we pass counts data to `scipp` as variance this is correct- but if one of the counts is 0 then its variance is 0, leading to a division by 0 error when calculating its weight for fitting `weight = 1 / doc["data"][self.yerr]`. See `src\ibex_bluesky_core\devices\dae\dae_spectra.py line 118`.
 
 ## Decision
 
--- a/manual_system_tests/dae_scan.py
+++ b/manual_system_tests/dae_scan.py
@@ -8,15 +8,17 @@ import bluesky.plan_stubs as bps
 import bluesky.plans as bp
 import matplotlib
 import matplotlib.pyplot as plt
-from bluesky.callbacks import LiveTable
+from bluesky.callbacks import LiveTable, LiveFitPlot
 from bluesky.preprocessors import subs_decorator
 from bluesky.utils import Msg
 from ophyd_async.plan_stubs import ensure_connected
 
 from ibex_bluesky_core.callbacks.file_logger import HumanReadableFileCallback
+from ibex_bluesky_core.callbacks.fitting import LiveFit
+from ibex_bluesky_core.callbacks.fitting.fitting_utils import Linear, Gaussian
 from ibex_bluesky_core.callbacks.plotting import LivePlot
 from ibex_bluesky_core.devices import get_pv_prefix
-from ibex_bluesky_core.devices.block import block_rw_rbv
+from ibex_bluesky_core.devices.block import block_rw_rbv, BlockWriteConfig
 from ibex_bluesky_core.devices.simpledae import SimpleDae
 from ibex_bluesky_core.devices.simpledae.controllers import (
     RunPerPointController,
@@ -47,10 +49,12 @@ def dae_scan_plan() -> Generator[Msg, None, None]:
     - The DAE waited for at least 500 good frames at each point
     """
     prefix = get_pv_prefix()
-    block = block_rw_rbv(float, "mot")
+    block = block_rw_rbv(float, "mot", write_config=BlockWriteConfig(settle_time_s=0.5))
+    blocka = block_rw_rbv(float, "alice")
+    blockb = block_rw_rbv(float, "bob", write_config=BlockWriteConfig(settle_time_s=0.5))
 
     controller = RunPerPointController(save_run=True)
-    waiter = GoodFramesWaiter(500)
+    waiter = GoodFramesWaiter(1)
     reducer = GoodFramesNormalizer(
         prefix=prefix,
         detector_spectra=[i for i in range(1, 100)],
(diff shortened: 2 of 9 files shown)


doc/architectural_decisions/005-variance-addition.md
```diff
@@ -0,0 +1,17 @@
+# Why do we add 0.5 to variance?
```
Tom-Willemsen, line 1: This probably needs a bit more expansion.

Here's what I'm thinking:

---

# Variance addition to counts data

## Status

Current

## Context

For counts data, the uncertainty on counts is typically defined by poisson counting statistics, i.e. the standard deviation on `N` counts is `sqrt(N)`.

This can be problematic in cases where zero counts have been collected, as the standard deviation will then be zero, which will subsequently lead to "infinite" point weightings in downstream fitting routines for example.

A number of possible approaches were considered:

| Option | Description |
| --- | --- |
| A | Reject data with zero counts, i.e. explicitly throw an exception if any data with zero counts is seen as part of a scan. |
| B | Use a standard deviation of `NaN` for points with zero counts. |
| C | Define the standard deviation of `N` counts as `1` if counts are zero, otherwise `sqrt(N)`. This is one of the approaches available in mantid for example. |
| D | Define the standard deviation of `N` counts as `sqrt(N+0.5)` unconditionally - on the basis that "half a count" is smaller than the smallest possible actual measurement which can be taken. |
| E | No special handling, calculate std. dev. as `sqrt(N)`. |

For clarity, the following table shows the value and associated uncertainty for each option:

| Counts | Std. Dev. (A) | Std. Dev. (B) | Std. Dev. (C) | Std. Dev. (D) | Std. Dev. (E) |
| ------- | ------ | ------- | ------- | ------- | --- |
| 0 | raise exception | NaN | 1 | 0.707 | 0 |
| 1 | 1 | 1 | 1 | 1.224745 | 1 |
| 2 | 1.414214 | 1.414214 | 1.414214 | 1.581139 | 1.414214 |
| 3 | 1.732051 | 1.732051 | 1.732051 | 1.870829 | 1.732051 |
| 4 | 2 | 2 | 2 | 2.12132 | 2 |
| 5 | 2.236068 | 2.236068 | 2.236068 | 2.345208 | 2.236068 |
| 10 | 3.162278 | 3.162278 | 3.162278 | 3.24037 | 3.162278 |
| 50 | 7.071068 | 7.071068 | 7.071068 | 7.106335 | 7.071068 |
| 100 | 10 | 10 | 10 | 10.02497 | 10 |
| 500 | 22.36068 | 22.36068 | 22.36068 | 22.37186 | 22.36068 |
| 1000 | 31.62278 | 31.62278 | 31.62278 | 31.63068 | 31.62278 |
| 5000 | 70.71068 | 70.71068 | 70.71068 | 70.71421 | 70.71068 |
| 10000 | 100 | 100 | 100 | 100.0025 | 100 |

## Present

These approaches were discussed in a regular project update meeting including
- TW & FA (Experiment controls)
- CK (Reflectometry)
- JL (Muons)
- RD (SANS)

## Decision

The consensus was to go with Option D.

## Justification

- Option A will cause real-life scans to crash in low counts regions.
- Option B involves `NaN`s, which have many surprising floating-point characteristics and are highly likely to be a source of future bugs.
- Option D was preferred to option C by scientists present.
- Option E causes surprising results and/or crashes downstream, for example fitting may consider points with zero uncertainty to have "infinite" weight, therefore effectively disregarding all other data.

doc/fitting/fitting.md
```diff
@@ -24,14 +24,17 @@ plt.figure()
 ax = plt.gca() 
 # ax is shared by fit_callback and plot_callback 
 
-plot_callback = LivePlot(y="y_variable", x="x_variable", ax=ax)
-fit_callback = LiveFit(Gaussian.fit(), y="y_variable", x="x_variable", update_every=0.5)
+plot_callback = LivePlot(y="y_signal", x="x_signal", ax=ax, yerr="yerr_signal_")
+fit_callback = LiveFit(Gaussian.fit(), y="y_signal", x="x_signal", yerr="yerr_signal_", update_every=0.5)
```
Tom-Willemsen, line 28: Should: Is the trailing underscore a typo? I don't think we would normally have any signals with trailing underscores?

doc/fitting/fitting.md
```diff
@@ -24,14 +24,17 @@ plt.figure()
...
+# Using the yerr parameter allows you to use error bars.
 # update_every = in seconds, how often to recompute the fit. If `None`, do not compute until the end. Default is 1.
 fit_plot_callback = LiveFitPlot(fit_callback, ax=ax, color="r")
 ```
 
 **Note:** that the `LiveFit` callback doesn't directly do the plotting, it will return function parameters of the model its trying to fit to; a `LiveFit` object must be passed to `LiveFitPlot` which can then be subscribed to the `RunEngine`. See the [Bluesky Documentation](https://blueskyproject.io/bluesky/main/callbacks.html#livefitplot) for information on the various arguments that can be passed to the `LiveFitPlot` class.
 
+Using the `yerr` argument allows you to pass uncertainties via a signal to LiveFit, so that the "weight" of each point influences the fit produced.
```
Tom-Willemsen, line 36: Could: Can we expand this slightly to say that the weight of each point is computed as `1/(standard deviation)`.

manual_system_tests/dae_scan.py
```diff
@@ -96,8 +113,9 @@ def dae_scan_plan() -> Generator[Msg, None, None]:
         ]
     )
     def _inner() -> Generator[Msg, None, None]:
-        num_points = 3
-        yield from bps.mv(dae.number_of_periods, num_points)
+        num_points = NUM_POINTS
```
Tom-Willemsen, line 116: Should: This seems a bit redundant, why did we need the duplicated variable?

src/ibex_bluesky_core/callbacks/fitting/__init__.py
```diff
@@ -50,31 +52,48 @@ class LiveFit(_DefaultLiveFit):
...
         """Call Bluesky LiveFit with assumption that there is only one independant variable.
 
         Args:
             method (FitMethod): The FitMethod (Model & Guess) to use when fitting.
             y (str): The name of the dependant variable.
             x (str): The name of the independant variable.
-            update_every (int): How often to update the fit. (seconds)
+            update_every (int, optional): How often to update the fit. (seconds)
+            yerr (str or None, optional): Name of field in the Event document
+            that provides standard deviation for each Y value
```
Tom-Willemsen, line 65: Could

[...], or None to not use fit weighting.

src/ibex_bluesky_core/devices/dae/dae_spectra.py
```diff
@@ -107,7 +109,14 @@ async def read_spectrum_dataarray(self) -> sc.DataArray:
         if unit is None:
             raise ValueError("Could not determine engineering units of tof edges.")
 
+        # TODO add reference to ADR
```
Tom-Willemsen, line 112: must: ☝️

tests/callbacks/test_plotting_callback.py
```diff
@@ -42,3 +44,56 @@ def test_show_plot_only_shows_if_backend_is_genie():
...
+    _, ax = plt.subplots()
+    ax.errorbar = MagicMock()
+
+    lp = LivePlot(y="y", x="x", ax=ax)
+
+    lp.start(
+        {
+            "time": 0,
+            "plan_name": "scan",
+            "uid": 0,  # type: ignore
```
Tom-Willemsen, line 93: Should: Isn't `uid` a str, I guess that's what pyright was complaining about?

tests/callbacks/test_plotting_callback.py
```diff
@@ -42,3 +44,56 @@ def test_show_plot_only_shows_if_backend_is_genie():
...
+            "scan_id": 0,
+        }
+    )
+
+    lp.update_plot()
+    assert not lp.ax.errorbar.called  # type: ignore
```
Tom-Willemsen, line 99: Could: I think if you make this just `ax.errorbar` you probably don't need the type ignore?

src/ibex_bluesky_core/callbacks/plotting.py
```diff
@@ -15,18 +16,56 @@
...
+        if yerr is not None:
+            self.yerr, *others = get_obj_fields([yerr])
+        else:
+            self.yerr = None
+        self.yerr_data = []
+
     def _show_plot(self) -> None:
+        """Call plt.show()."""
         # Play nicely with the "normal" backends too - only force show if we're
         # actually using our custom backend.
```
Tom-Willemsen, line 47: Could: can you merge the existing comment into the new docstring?

Tom-Willemsen (commented):
Could: The docs changes are quite minimal... Is that really all we need to say about this change?

Must: not your fault but this has a merge conflict with `main` now, can you merge in `main`?

tests/callbacks/test_plotting_callback.py
jackbdoughty: Cannot access attribute "ax" for class "LivePlot". Might have to leave this one

Tom-Willemsen (approved):

PR #54, opened 2024-11-07
add logo to README

rerpha:


Tom-Willemsen:
I was just reading about this, apparently to get it to work on our pypi front page we'll need to use an absolute link.

Not tried but think `` should work?

Tom-Willemsen (approved):

rerpha:
I approve

PR #56, opened 2024-11-11
add action for checking every day against ophyd-async and bluesky main branches

### LINKED ISSUE #49: Periodic CI against `main` of `bluesky` and `ophyd-async`
As a developer I would like an early preview if upstream changes to `bluesky` or `ophyd-async` might cause problems for us.

Run a periodic (nightly/weekly?) CI job which explicitly installs the `main` branch of both `bluesky` and `ophyd-async` (as opposed to released versions which we typically use), and run:
- `ruff/pyright`
- `pytest`

rerpha:
may close #49 

this is difficult to test as [the action will only work on the main branch](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule):
> This event will only trigger a workflow run if the workflow file is on the default branch.

### DIFF
--- a/.github/workflows/test-against-main.yml
+++ b/.github/workflows/test-against-main.yml
@@ -3,11 +3,6 @@ on:
   schedule:
     - cron: "0 0 * * *"
 jobs:
-  call-linter-workflow:
-    uses: ISISComputingGroup/reusable-workflows/.github/workflows/linters.yml@main
-    with:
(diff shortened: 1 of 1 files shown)


Tom-Willemsen (approved):
Can't really tell whether this will work or not without trying it :)

PR #63, opened 2024-11-14
rename dampedosc.PNG to dampedosc.png to fix sphinx build

rerpha:


rerpha:
too slow...

Tom-Willemsen:
Ha sorry well you were 30s faster on the `reusable-workflows` one so I had to be faster on this one ;)

Issue #72, opened 2024-11-22
Slit scan guess


jackbdoughty:
Slit scan guess was sometimes strugglnig to converge due to initial guesses outside of the data, this causes all differentials to go to zero for some parameters and hence optimizer cannot work.

Simplified heuristic guessing function that keeps all initial parameters within the bounds of the data seemed to work well to give the optimizer a good chance of converging to a decent fit:

```
            background = np.min(y)
            inflection0 = np.min(x) + (1/3) * (np.max(x) - np.min(x))
            inflections_diff = (1/3) * (np.max(x) - np.min(x))
            gradient = 2 * (np.max(y) - np.min(y)) / (np.max(x) - np.min(x))
            height_above_inflection1 = (np.max(y) - np.min(y)) / 5.0
```

Needs unit tests etc adding

jackbdoughty:
See https://github.com/ISISComputingGroup/ibex_bluesky_core/pull/76

PR #76, opened 2024-11-22
fixes slit scan guess & unit tests

jackbdoughty:
Slit scan guess was sometimes struggling to converge due to initial guesses outside of the data, this causes all differentials to go to zero for some parameters and hence optimizer cannot work.

Simplified heuristic guessing function that keeps all initial parameters within the bounds of the data seemed to work well to give the optimizer a good chance of converging to a decent fit.

Updates unit tests.

To test make sure that all unit tests pass.


jackbdoughty:
Original issue: 
https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/72

Tom-Willemsen (commented):
There's a table in the docs [somewhere ](https://isiscomputinggroup.github.io/ibex_bluesky_core/fitting/fitting.html#models)that says you can pass a min width to `SlitScan`, that's no longer true...

(Maybe we could _make_ it true again in future by having a min-slit-width as a percentage of the total x-range, that would potentially let us guarantee again that the guess in within the range of the data... but well out of scope for this I think)

Tom-Willemsen (approved):

Issue #69, opened 2024-11-22
`init guess` label not appearing in legend of `LiveFitPlot`


jackbdoughty:
It previously did appear but somehow seems to have got lost.

Should look like this (example from bluesky docs):

![image](https://github.com/user-attachments/assets/ae28dc03-72e9-4473-8d39-49d5e97055e2)

Figure out when we accidentally disabled it and undo that.

Maybe also play with different colours or perhaps dashed lines etc to see if we can further disambiguate the guess more on the plot? But the guess is liked and still wanted on the plot.

PR #79, opened 2024-11-24
Force legend to be drawn

Tom-Willemsen:
Fixes https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/69

Happens if a fit guess doesn't have enough parameters on first point to generate a line.


Tom-Willemsen:
Hmm not sure this is right actually.

PR #83, opened 2024-11-25
Set uncertainties more correctly

Tom-Willemsen:
Invalid normalization should crash a scan - there's really no way around this, we should not "pretend" to be able to do it. Have checked with scientists that this makes sense to them and they agreed an exception is better than giving "some" data which would be meaningless. In real life, monitor counts being zero means something is fundamentally wrong with the beamline (shutter closed, monitor not in, etc) so the scan can never really be valid in that state.

Also move the variance addition later - after summing all the detectors - basically do it at the last level possible when our units are still "counts". This avoids broadcasting the "extra" variance across multiple tof channels / pixels.


src/ibex_bluesky_core/callbacks/reflectometry/_det_map.py
```diff
@@ -60,23 +61,23 @@ def event(self, doc: Event, **kwargs: dict[str, Any]) -> Event:
...
-            dims=["spectrum"], values=mon_data, variances=mon_data + 0.5, dtype="float64"
-        )
+        det = sc.Variable(dims=["spectrum"], values=det_data, variances=det_data, dtype="float64")
+        mon = sc.Variable(dims=["spectrum"], values=mon_data, variances=mon_data, dtype="float64")
 
         det_sum = det.sum()
         mon_sum = mon.sum()
 
+        # See doc\architectural_decisions\005-variance-addition.md
+        # for justfication of this addition to variances.
```
rerpha, line 71: minor nitpick but as its in a few places, `justfication` is a typo

rerpha (approved):

PR #88, opened 2024-11-29
Add option to wait for global moving flag

Tom-Willemsen:
Closes https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/77

### DIFF
--- a/doc/devices/blocks.md
+++ b/doc/devices/blocks.md
@@ -131,6 +131,10 @@ from ibex_bluesky_core.devices.block import block_mot
 mot_block = block_mot("motor_block")
 ```
 
+A motor block does not need an explicit write config: it always waits for the requested motion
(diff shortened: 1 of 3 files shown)


doc/devices/blocks.md
```diff
@@ -136,7 +136,7 @@ mot_block = block_mot("motor_block")
 `BlockRw` and `BlockRwRbv` both take a `write_config` argument, which can be used to configure
```
rerpha, line 136: (I can't comment on a few lines above but this is referring to the block_mot bit above) 

we should probably just mirror what the docstring says about motors not explicitly needing a wait on their own underneath the "run control" bit as `waitfor_move` is pretty commonly used in the scans library I think.

rerpha (approved):

Issue #24, opened 2024-08-22
Scan IDs, filenames, saved files, rb numbers, archiving, etc


Tom-Willemsen:
As a data administrator or instrument scientist I would like to figure out how scans output data gets stored and named.

`bluesky` output files contain scientifically-useful data, grouping together a series of related measurements into a coherent "dataset".

@GRyall and @ChrisM-S have already both told me they have opinions on this topic.

## Acceptance criteria

Documented resolution (and implementation tickets) for the following:
- [x] Figure out what our file naming scheme is for bluesky scans, whether we use GUID or `scan_id` in various combinations with/without instrument name prepended and so on.
- [x]  Figure out where `bluesky` can write data so that it is treated as long-term scientifically-useful data and gets archived in the right ways for long-term scientifically-useful data.
- [x] Do downstream systems need to associate these files with an RB number? How does that work and what metadata do we need to add to our files to help downstream systems?

## Notes
- **A bluesky "scan" has no defined relationship to a DAE run**. 
  * It may contain zero or one or many DAE runs.
- It is reasonable to assume that a scan will sit within a single RB number and a single cycle.
- Bluesky has the concept of a `scan_id` - a numeric ID. By default this is not persisted anywhere and simply starts from 1 and increments in each new python session, but it is easy to configure the run engine to get the "next scan id" from any source (e.g. a file, database, network service, etc etc). If we want a constantly-incrementing scan id on a per-instrument basis that is easy to do.
- Bluesky _also_ assigns a globally-unique ID to each scan. 
  * NSLS2 use GUIDs for all scan data and don't use a `scan_id` at all, as far as I understand it.

## Planning
Time in recording of planning meeting: 01h15m 2024/11/28
Also discussed 19/12/2024

jackbdoughty:
Reflectometry scientists, at least, would like `axis_date_time` formatted filenames - therefore not using scan_ids.

They would like them in a cycle-number specific folder (the cycle number is available as a PV from the DAE)

Tom-Willemsen:
Bring this up in discussion in one of the scans catchup meetings, try to get all science groups to vaguely agree with one another as step 1.

The _directory_ into which files should be written almost certainly needs to be configurable per-instrument (i.e. set in each instrument's plan wrappers). Hopefully we can agree on a consistent scheme for the file names themselves.

LowriJenkins:
Resolutions pending further discussion with scientists:
-  `axis_date_time` format for filenames
[...]

PR #98, opened 2025-02-03
fix exception if output file (rbnum) dir doesn't exist

### LINKED ISSUE #99: human readable files: all fields put in rather than just measured fields
#24 introduced a new format in which the human-readable files are output as `instrument_axis_date_time` but currently the writer just puts all fields specified as the `axis` part rather than _just_ the measured fields. 

You can get these from the start message from `"detectors"` and `"motors"` (sometimes, we should account for there being no motors as `count()` for example won't put them in the start message)

rerpha:
this fixes an exception that occurs when the HumanReadableFileCallback tries to create a file if the parent directory does not exist. 

Closes #99

### DIFF
--- a/src/ibex_bluesky_core/callbacks/file_logger.py
+++ b/src/ibex_bluesky_core/callbacks/file_logger.py
@@ -62,17 +62,13 @@ class HumanReadableFileCallback(CallbackBase):
         )
         rb_num = doc.get("rb_number", "Unknown RB")
 
-        detectors = doc.get("detectors", [])
         # motors is a tuple, we need to convert to a list to join the two below
-        motors = doc.get("motors", [])
-        if motors:
-            motors = list(motors)
+        motors = list(doc.get("motors", []))
 
-        detectors_and_motors = "_".join(detectors + motors)
         self.filename = (
             self.output_dir
             / f"{rb_num}"
-            / f"{INSTRUMENT}_{detectors_and_motors}_{title_format_datetime}Z.txt"
+            / f"{INSTRUMENT}{'_' + '_'.join(motors) if motors else ''}_{title_format_datetime}Z.txt"
         )
         if rb_num == "Unknown RB":
             logger.warning('No RB number found, saving to "Unknown RB"')
@@ -84,9 +80,7 @@ class HumanReadableFileCallback(CallbackBase):
         ]
         header_data = {k: v for k, v in doc.items() if k not in exclude_list}
 
-        formatted_time = datetime_obj.astimezone(ZoneInfo("Europe/London")).strftime(
-            "%Y-%m-%d %H:%M:%S"
-        )
+        formatted_time = datetime_obj.astimezone(ZoneInfo("UTC")).strftime("%Y-%m-%d %H:%M:%S")
         header_data[START_TIME] = formatted_time
 
         # make sure the parent directory exists, create it if not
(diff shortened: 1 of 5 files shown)


src/ibex_bluesky_core/callbacks/file_logger.py
```diff
@@ -79,9 +79,11 @@ def start(self, doc: RunStart) -> None:
         )
         header_data[START_TIME] = formatted_time
 
+        if not self.filename.parent.exists():
+            self.filename.parent.mkdir()
```
Tom-Willemsen, line 83: Usually ` os.makedirs(name, exist_ok=True)` is recommended as:
- It makes intermediate directories
- It's less likely to suffer from race conditions between the check for dir existing and creation

src/ibex_bluesky_core/callbacks/file_logger.py
```diff
@@ -79,9 +79,11 @@ def start(self, doc: RunStart) -> None:
...
+
         with open(self.filename, "a", newline="", encoding="utf-8") as outfile:
-            for key, value in header_data.items():
-                outfile.write(f"{key}: {value}\n")
+            outfile.writelines([f"{key}: {value}\n" for key, value in header_data.items()])
```
Tom-Willemsen, line 86: If using `writelines` do we actually need a trailing `\n`? I haven't checked but doesn't this give us double terminators?

Tom-Willemsen: Ah no I see we're passing `newline=""` to `open()` - my bad...

src/ibex_bluesky_core/callbacks/file_logger.py
```diff
@@ -59,10 +60,19 @@ def start(self, doc: RunStart) -> None:
...
-        axes = "_".join(self.fields)
         rb_num = doc.get("rb_number", "Unknown RB")
+
+        detectors = doc.get("detectors", [])
+        # motors is a tuple, we need to convert to a list to join the two below
+        motors = doc.get("motors", [])
+        if motors:
+            motors = list(motors)
+
+        detectors_and_motors = "_".join(detectors + motors)
```
Tom-Willemsen, line 71: My interpretation of https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/24#issuecomment-2590275747 was that `instrument_axis_date_time` means `axis` is just the motors - explicitly _not_ all the detectors (of which there may be very many, whereas in pretty much all cases there'll only be one or _maybe_ a few motors.

i.e. for a scan of `theta` against `dae-intensity` and `dae-intensity_err` I think the filename should be:

```
CRISP_theta_2024_01_01-01_01_01Z.csv
```

Tom-Willemsen (commented):
sorry I keep adding comments but just noticed a couple more things which would be useful to sort while we're here...

src/ibex_bluesky_core/callbacks/file_logger.py
```diff
@@ -59,10 +60,19 @@ def start(self, doc: RunStart) -> None:
...
         self.filename = (
-            self.output_dir / f"{rb_num}" / f"{INSTRUMENT}_{axes}_{title_format_datetime}Z.txt"
+            self.output_dir
+            / f"{rb_num}"
+            / f"{INSTRUMENT}_{detectors_and_motors}_{title_format_datetime}Z.txt"
```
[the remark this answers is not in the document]
rerpha: makes sense - will use UTC. i think we should use `.txt` as these files will probably break excel (or any other csv reader) if someone opens them up in it...

Tom-Willemsen (dismissed):

Tom-Willemsen (approved):

PR #104, opened 2025-02-10
Isis standard cb

rerpha:


manual_system_tests/dae_scan.py
```diff
@@ -73,66 +67,37 @@ def dae_scan_plan() -> Generator[Msg, None, None]:
...
+    icc = ISISCallbacks(
+        x=block.name,
+        y=reducer.intensity.name,
+        yerr=reducer.intensity_stddev.name,
+        fit=Linear.fit(),
+        measured_fields=[
+            block.name,
+            controller.run_number.name,
+            reducer.intensity.name,
+            reducer.intensity_stddev.name,
```
Tom-Willemsen, line 81: I think we should assume that x, y, (y_err) if provided are measured - so this should be _extra_ measured fields in addition to those 2/3 I think

src/ibex_bluesky_core/callbacks/__init__.py
```diff
@@ -1 +1,203 @@
...
+                    linestyle="none",
+                    ax=ax,
+                    yerr=yerr,
+                )
+            )
+
+    @property
+    def live_fit(self) -> LiveFit | None:
+        """The live fit object containing fitting results."""
+        return self._live_fit
```
Tom-Willemsen, line 187: I think for pyright we want this to be `-> LiveFit` and inside the implementation throw a `ValueError` or something if it's None.

Otherwise people have to do:

```
the_fit = icc.live_fit
assert the_fit is not None
print(the_fit.fit_report())
```
to placate pyright downstream

src/ibex_bluesky_core/callbacks/__init__.py
```diff
@@ -1 +1,203 @@
...
+            cb = _Cb()
+            cb("start", {"time": 0, "uid": ""})
+            done_event.wait(10.0)
+
+        if add_fit_cb:
+            if fit is None:
+                raise ValueError("fit method must be specified if add_fit_cb is True")
+            self._live_fit = LiveFit(fit, y=y, x=x, yerr=yerr)  # pyright: ignore reportArgumentType
+            if show_fit_on_plot:
+                self.subs.append(LiveFitPlot(livefit=self._live_fit, ax=ax))
```
Tom-Willemsen, line 170:
```suggestion
            if show_fit_on_plot:
                self.subs.append(LiveFitPlot(livefit=self._live_fit, ax=ax))
            else:
                self.subs.append(self._live_fit)
```

Otherwise I can't see that fit ever gets subscribed?

src/ibex_bluesky_core/callbacks/__init__.py
```diff
@@ -1 +1,203 @@
...
+                    output_dir=Path(human_readable_file_output_dir)
+                    if human_readable_file_output_dir
+                    else DEFAULT_PATH,
+                ),
+            )
+
+        if add_table_cb:
+            combined_lt_fields = measured_fields + fields_for_live_table
+            if not combined_lt_fields:
+                raise ValueError("No fields specified for the live table")
```
Tom-Willemsen, line 138: I think we mandate at least `x` and `y` for this callback collection, and then this can never occur

src/ibex_bluesky_core/callbacks/__init__.py
```diff
@@ -1 +1,240 @@
...
+            measured_fields = []
+        if fields_for_live_table is None:
+            fields_for_live_table = []
+        if fields_for_hr_file is None:
+            fields_for_hr_file = []
+
+        measured_fields.append(x)
+        measured_fields.append(y)
+
+        if show_fit_on_plot and (not add_fit_cb or fit is None):
```
jackbdoughty, line 134: This might be a bit petty- but would this not mean that by default, if you do not provide anything for `fit` or `show_fit_on_plot` arguments then this will raise an error. Would it be better to have something that says if you don't provide a `fit` then this is off by default but if you do then this is on by default?

src/ibex_bluesky_core/callbacks/__init__.py
```diff
@@ -1 +1,240 @@
...
+        self,
+        x: str,
+        y: str,
+        yerr: str | None = None,
+        measured_fields: list[str] | None = None,
+        add_table_cb: bool = True,
+        fields_for_live_table: list[str] | None = None,
+        add_human_readable_file_cb: bool = True,
+        fields_for_hr_file: list[str] | None = None,
+        human_readable_file_output_dir: str | PathLike[str] | None = None,
```
jackbdoughty, line 52: Another petty-ish thing- the type signatures for arguments `human_readable_file_output_dir` and `live_fit_logger_output_dir` are slightly different but they're both paths. It makes it a bit confusing as you can put a str in for  `human_readable_file_output_dir` but `live_fit_logger_output_dir` won't allow the same.

jackbdoughty (changes requested):
Looks good to me, just these two things I think.

src/ibex_bluesky_core/callbacks/__init__.py
```diff
@@ -1 +1,240 @@
...
+logger = logging.getLogger(__name__)
+
+# ruff: noqa: PLR0913, PLR0912, PLR0917
+
+
+class ISISCallbacks:
+    """ISIS standard callbacks for use within plans."""
+
+    def __init__(
+        self,
```
rerpha, line 43: make kwarg only with * first

jackbdoughty (approved):

PR #105, opened 2025-02-10
78 docs make classfunction names link through to api reference more consistently

danielmaclaren:


doc/devices/blocks.md
```diff
@@ -132,20 +132,20 @@ mot_block = block_mot("motor_block")
...
 
-See {py:class}`ibex_bluesky_core.devices.block.BlockWriteConfig` for a detailed
+See [`BlockWriteConfig`](ibex_bluesky_core.devices.block.BlockWriteConfig) for a detailed
 description of all the options which are available.
 
-## Run control
+## Run Control
 
-Run control information is available via the `block.run_control` sub-device.
+[`RunControl`](ibex_bluesky_core.devices.block.RunControl) information is available via the [`block.run_control`](ibex_bluesky_core.devices.block.run_control) sub-device.
```
jackbdoughty, line 148: Sphynx isn't very happy with some of these python references, I think for this one you want the same again: `ibex_bluesky_core.devices.block.RunControl` as I don't think there is a `ibex_bluesky_core.devices.block.run_control`. The checks should pass when it can find all the references

jackbdoughty (changes requested):
Looks like this is almost done- there's a stray `test.py` file with nothing in it in this PR, which ruff is complaining about- is this meant to be part of it? Other than that just the sphynx errors mentioned below

doc/devices/blocks.md
danielmaclaren: Sorry about that, updated it and hopefully it passes!

danielmaclaren: There's a few other links that need updating, I'll fix them and let you know once sphinx passes

danielmaclaren:
@jackbdoughty seems that the tests have passed now

jackbdoughty (approved):

PR #106, opened 2025-02-10
Adds our own CoM callback

jackbdoughty:
We could not use Bluesky's centre of mass callback as it was not suited for the way we would want to use it, so this is our implementation for one.

To test:

- Make sure that all tests pass.
- Run the following scans and check that you get sensible CoMs for each. (Send me a message if you need to know how to set up to scan blocks against eachother / create fake data):
  - A scan with non constant point spacing.
  - A scan where some of the values are negative / have a background.
  - The "there-and-back" scan so x points would be like: -2 -1 0 1 2 1 0 -1 -2.  (Non continuous points)

### DIFF
--- a/doc/architectural_decisions/006-peak-stats.md
+++ b/doc/architectural_decisions/006-peak-stats.md
@@ -1,0 +1,21 @@
+# Using our own Centre of Mass Callback
+
+## Status
+
+Current
+
+## Context
+
+A decision needs to be made about whether to make changes to upstream Bluesky so that their `CoM` callback works for us, or we make our own.
+
+## Decision
+
+We will be making our own `CoM` callback.
+
+## Justification & Consequences
+
+We attempted to make changes to upstream Bluesky which were rejected, as it adds limits to the functionality of the callback. We also found other limitations with using their callback, such as not being able to have disorderd and non-continuous data sent to it without it skewing the calculated value- we need it to work with disordered and non-continuous data as we need to be able to run continuous scans.
+
+This will mean that...
+- Our version of the callback will not be supported by Bluesky and may need changes as Bluesky updates.
+- We can have a version of the callback that is made bespokely for our use cases.
--- a/doc/architectural_decisions/006-peak-stats.md
+++ b/doc/architectural_decisions/006-peak-stats.md
@@ -14,7 +14,7 @@ We will be making our own `CoM` callback.
 
 ## Justification & Consequences
 
-We attempted to make changes to upstream Bluesky which were rejected, as it adds limits to the functionality of the callback. We also found other limitations with using their callback, such as not being able to have disorderd and non-continuous data sent to it without it skewing the calculated value- we need it to work with disordered and non-continuous data as we need to be able to run continuous scans.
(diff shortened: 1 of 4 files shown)


doc/architectural_decisions/006-peak-stats.md
```diff
@@ -0,0 +1,21 @@
...
+
+A decision needs to be made about whether to make changes to upstream Bluesky so that their `CoM` callback works for us, or we make our own.
+
+## Decision
+
+We will be making our own `CoM` callback.
+
+## Justification & Consequences
+
+We attempted to make changes to upstream Bluesky which were rejected, as it adds limits to the functionality of the callback. We also found other limitations with using their callback, such as not being able to have disorderd and non-continuous data sent to it without it skewing the calculated value- we need it to work with disordered and non-continuous data as we need to be able to run continuous scans.
```
rerpha, line 17:
```suggestion
We attempted to make changes to upstream Bluesky which were rejected, as it adds limits to the functionality of the callback. We also found other limitations with using their callback, such as not being able to have disordered and non-continuous data sent to it without it skewing the calculated value- we need it to work with disordered and non-continuous data as we need to be able to run continuous scans.
```

src/ibex_bluesky_core/callbacks/fitting/__init__.py
```diff
@@ -127,3 +129,62 @@ def update_fit(self) -> None:
...
+
+    sum_xyw = np.sum(x_sorted * y_sorted * weight)  # Weighted CoM calculation
+    sum_yw = np.sum(y_sorted * weight)
+    com_x = sum_xyw / sum_yw
+
+    return com_x
+
+
+@make_class_safe(logger=logger)  # pyright: ignore (pyright doesn't understand this decorator)
+class PeakStats(_DefaultPeakStats):
```
rerpha, line 173: we should use this in our standard callbacks class - if this ticket is merged before #104 i will add it in, if not i think this PR should

jackbdoughty:
https://github.com/ISISComputingGroup/ibex_bluesky_core/pull/108 Will fix the above errors

src/ibex_bluesky_core/callbacks/fitting/__init__.py
```diff
@@ -127,3 +129,57 @@ def update_fit(self) -> None:
...
```
Tom-Willemsen, line 168: Sorry I know I've gone back and forth on this, but I'm not fully convinced it's a good idea to subclass `_DefaultPeakStats` here actually - the upstream one has `__slots__` defined and some other potential traps.

Maybe here it is better to just make a standalone callback... subclassing `CollectThenCompute` or similar.

Tom-Willemsen (commented):
I'm going to push some stuff to a `trapeze` branch and we should probably chat tomorrow... I found some extra-fun edge cases that I hadn't previously considered...

e.g. this case:

```python
def test_tricky_non_constant_spacing():
    ps.start({})

    ps.event({"data": {"x": 0, "y": 1}})
    ps.event({"data": {"x": 0.1, "y": 1}})
    ps.event({"data": {"x": 4, "y": 1}})
    ps.event({"data": {"x": 5, "y": 0}})
    ps.event({"data": {"x": 6, "y": 1}})
    ps.event({"data": {"x": 10, "y": 1}})

    ps.stop({})

    assert ps["com"] == 5.0
```

Issue #68, opened 2024-11-22
Consolidate import name spaces


jackbdoughty:
- As a developer I would like the import namespaces to be a little leaner so that I need fewer lines of imports
- As a scientist I would like imports such as `fitting_utils` to have more concise names

For the test on CRISP our imports ended up looking like this, for relatively simple scans:

```
from ibex_bluesky_core.callbacks.file_logger import HumanReadableFileCallback
from ibex_bluesky_core.callbacks.fitting import LiveFit
from ibex_bluesky_core.callbacks.fitting.livefit_logger import LiveFitLogger
from ibex_bluesky_core.callbacks.fitting.fitting_utils import Fit, Linear
from ibex_bluesky_core.callbacks.plotting import LivePlot
from ibex_bluesky_core.devices import get_pv_prefix
from ibex_bluesky_core.devices.block import BlockRw, BlockWriteConfig, block_rw
from ibex_bluesky_core.devices.simpledae import SimpleDae
from ibex_bluesky_core.devices.simpledae.controllers import (
    PeriodPerPointController,
    RunPerPointController,
)
from ibex_bluesky_core.devices.simpledae.reducers import MonitorNormalizer
from ibex_bluesky_core.devices.simpledae.waiters import GoodFramesWaiter, PeriodGoodFramesWaiter
from ibex_bluesky_core.run_engine import get_run_engine
from ibex_bluesky_core.callbacks.fitting import fitting_utils as fit
```

This is a bit verbose, mostly due to having to import things from multiple levels in a hierarchy

### Acceptance criteria
- [ ] Go through and add relevant public imports to `__init__.py`s and `__all__` so that they are accessible from "higher level" modules - e.g. all `simpledae` functionality should probably be accessible from `ibex_bluesky_core.devices.simpledae`.
- [ ] We may consider making more of our modules `_private`, so that it's clear what are the user-facing import paths and what is implementation details of the library

### Notes
- We do not need to change our _internal_ structure to do this - just expose things via imports and `__all__` at the right levels.

# Planning
10/01/25 - 00:40:50

Tom-Willemsen:
Could also look at flattening out some of our classes a bit - e.g. could `Gaussian().fit()` be just `gaussian` as a free function?

jackbdoughty:
For whoever picks this up, make sure to update `doc\fitting\fitting.md` according to new class/function/import names

rerpha:
[...]

PR #111, opened 2025-02-18
Add plans

### LINKED ISSUE #109: add plans area
we are going to be in charge of plans moving forward, so we should add a "plans" area for keeping track of these. this has the benefit that they'll get picked up by our CI (as our manual system tests do)

We should lay them out in a hierarchical structure ie. 

```
ibex_bluesky_core/
  plans/
    reflectometry/
        inter/
        polref/
    sans/
        loq/
        sans2d/
```

etc. 

devices should be kept in the devices module, though if we need to add instrument-specific we should follow the above. there will be devices specific to groups as well and so on. 

we have some plans and devicse we've used during testing, so adding these would be a good start.

rerpha:
Closes #109 

This PR adds some plans designed to help users out - we decided on doing this in #114 

The logic of these are mostly in the base plans which take devices rather than strings as arguments, which in turn are just thin wrappers around the various scan plans that bluesky provides. 

I've tried to document the above in docs so it's clear to users. 

I think by providing these we are doing a part of #68 as _most_ plans will probably use these and the callbacks/dae objects are going to be covered by the stuff in this PR which will reduce imports. we should probably still consolidate the imports further and/or move to a structure more similar to ophyd-async where we have things exposed in `__init__.py` but have files prefixed with `_` to indicate they shouldn't be used outside of this library


doc/plans/plans.md
```diff
@@ -0,0 +1,41 @@
+# Plans 
+
+`ibex_bluesky_core` provides a number of plans that can be run by the RunEngine directly. 
+
+These take "devices" so you can pass your own DAE object and a movable/readable: 
+
+[`scan`](ibex_bluesky_core.plans.scan) - this is for a normal/relative scan.
```
Tom-Willemsen, line 7:
```suggestion
[`scan`](ibex_bluesky_core.plans.scan) - this is for a absolute/relative scan.
```

doc/plans/plans.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Technique-specific plans
+
+### Reflectometry
+
+[`refl_scan`](ibex_bluesky_core.plans.reflectometry.refl_scan)
+
+[`refl_adaptive_scan`](ibex_bluesky_core.plans.reflectometry.refl_adaptive_scan)
+
+These are very similar to the above plans but are designed to construct a DAE and a [`ReflParameter`](ibex_bluesky_core.devices.reflectometry.ReflParameter) given a reflectometry server parameter name. These have some logic in the reflectometry server which tell us if sets were successful and so on.
```
Tom-Willemsen, line 40: > and so on

I think we probably need to be a bit more precise in the docs?

src/ibex_bluesky_core/devices/reflectometry/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+
+        """
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.readback: SignalR[float] = epics_signal_r(float, f"{prefix}REFL_01:PARAM:{name}")
+        self.setpoint: SignalW[float] = epics_signal_w(float, f"{prefix}REFL_01:PARAM:{name}:SP")
+        self.changing: SignalR[bool] = epics_signal_r(
+            bool, f"{prefix}REFL_01:PARAM:{name}:CHANGING"
+        )
+        self.redefine = ReflParameterRedefine(
+            prefix=f"{prefix}REFL_01:PARAM:{name}:", name=name + "redefine"
```
Tom-Willemsen, line 37: Why do we need an explicit name here? Won't it get renamed by ophyd-async anyway?

src/ibex_bluesky_core/devices/reflectometry/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+        """Reflectometry server parameter redefinition.
+
+        Args:
+            prefix: the reflectometry parameter full address.
+            name: the name of the parameter redefinition.
+            changed_timeout: time to wait for the CHANGED signal to go to False after a set.
+
+        """
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.changed: SignalR[bool] = epics_signal_r(bool, f"{prefix}DEFINE_POS_CHANGED")
```
Tom-Willemsen, line 74: Not fully convinced by making `changed` hinted - this feels like a case where we might want to hint nothing?

Just generally what it'd roughly expect is that the hinted-signal should _read_ whatever `set()` _writes_, usually...

(we can chat about it if you want)

src/ibex_bluesky_core/devices/reflectometry/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+        self.define_pos_sp = epics_signal_w(float, f"{prefix}DEFINE_POS:SP")
+        self.changed_timeout = changed_timeout
+        super().__init__(name)
+
+    @AsyncStatus.wrap
+    async def trigger(self) -> None:  # noqa: D102
+        pass  # pragma: no cover
```
Tom-Willemsen, line 81: `BlockR`'s only need a `trigger()` so that adaptive scans can use them... but I can't see someone wanting to adaptive-scan redefining a refl parameter... Likely this can just go

src/ibex_bluesky_core/devices/reflectometry/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @AsyncStatus.wrap
+    async def set(self, value: float) -> None:
+        """Set the setpoint."""
```
Tom-Willemsen, line 85: I guess it's _technically_ true that it

> Sets the setpoint [...of the redefine PV]

But I feel this needs clarifying. This probably needs to state that it's redefining the position etc etc, with appropriate warnings/caveats about it's use.

src/ibex_bluesky_core/devices/reflectometry/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def refl_parameter(name: str, changing_timeout: float = 60.0) -> ReflParameter:
+    """Small wrapper around a reflectometry parameter device.
+
+    This automatically applies the current instrument's PV prefix.
+
+    Args:
+        name: the reflectometry parameter name.
+        changing_timeout: time to wait for the CHANGED signal to go to False after a set.
```
Tom-Willemsen, line 100:
```suggestion
        changing_timeout_s: time to wait for the CHANGED signal to go to False after a set.
```

or 

```suggestion
        changing_timeout: time to wait, in seconds, for the CHANGED signal to go to False after a set.
```

Tom-Willemsen, line 100:
```suggestion
        changing_timeout: time to wait for the CHANGING signal to go to False after a set.
```

Tom-Willemsen (commented):
apologies if i'm being a bit picky...

src/ibex_bluesky_core/devices/reflectometry/__init__.py
rerpha: it is actually the `CHANGED` signal for a redefine - helpfully the opposite to a normal parameter...

rerpha: actually no im being dumb sorry, looking at the wrong bit of code

src/ibex_bluesky_core/devices/reflectometry/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+        """
+        self.changed: SignalR[bool] = epics_signal_r(bool, f"{prefix}DEFINE_POS_CHANGED")
+        self.define_pos_sp = epics_signal_w(float, f"{prefix}DEFINE_POS:SP")
+        self.changed_timeout = changed_timeout_s
+        super().__init__(name)
+
+    @AsyncStatus.wrap
+    async def set(self, value: float) -> None:
+        """Set the setpoint.
```
Tom-Willemsen, line 84: I still think this should be more explicit that it's _redefining_ the position of the axis when this `set()` method gets called - I know it's in a `ReflParameterRedefine` device but given redefining can lead to all sorts of subtle stuff, I'd rather be explicit

doc/plans/plans.md
```diff
@@ -1,34 +1,85 @@
...
-`ibex_bluesky_core` provides a number of plans that can be run by the RunEngine directly. 
+`ibex_bluesky_core` provides a number of plans that can be run by the RunEngine directly. These are mostly thin wrappers around the generic [bluesky plans](https://blueskyproject.io/bluesky/main/plans.html)
+
+All of these plans can be used directly by the RunEngine (`RE()`) or within a wrapper of your own that yields from them (`yield from`)
+
+So you can do this: 
+
+```python
+
+    >>> icc = RE(yield from refl_scan("S1VG", 1, 10, 21, model=Gaussian().fit() frames=500, det=100, mon=3, pixel_range=6, periods=True, save_run=False))
```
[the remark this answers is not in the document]
rerpha: you're right, no idea why i put that.. think i was getting confused with icc = (yield from plan) 
i don't know if there's a way to use icc though after calling one of these plans from the run engine?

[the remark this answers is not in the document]
Tom-Willemsen: To get the results out of icc.fit_result for example?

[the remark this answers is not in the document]
rerpha: maybe i'll just remove that you can call these directly - i guess we want people to write their own inst-specific wrappers

doc/plans/plans.md
```diff
@@ -1,34 +1,85 @@
+from ibex_bluesky_core.plans import motor_scan
```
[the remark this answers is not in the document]
Tom-Willemsen: This is how it renders for me, is this what you intended?

image

doc/plans/plans.md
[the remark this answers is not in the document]
rerpha: wonder if i should leave the "passing to run engine directly" in but with the caveat that its essentially pointless as any output is lost

jackbdoughty:
Having a look now...

Issue #113, opened 2025-02-20
standard callbacks: livefitlogger postfix default doesn't make sense


rerpha:
currently it's `isc` - we should have a think about what would be best. perhaps just `fit`? 

default is defined here: https://github.com/ISISComputingGroup/ibex_bluesky_core/blob/main/src/ibex_bluesky_core/callbacks/__init__.py#L61

Issue #103, opened 2025-02-04
Reflectometry: detector-mapping sample alignment


Tom-Willemsen:
Implement DAE detector-mapping scans for POLREF.

Current workflow, as far as I understand it, is:
- Do a height scan, maximising integrated intensity across detector
- Do an angle scan, using central detector pixel as a "god pixel", maximising intensity on it.

This is slow due to needing two physical scans and needing to count for quite a while on each angle scan point to get decent statistics

---

Conceptually the detector-mapping alignment gives POLREF, and other reflectometers:
- A _simultaneous_ height & angle sample alignment scan - i.e. two scans for the price of one.
- More counts on the angle scan, due to accumulating data from many DAE runs and all detector pixels, which correspondingly reduces counting time required to get viable statistics on the angle scan

Together these two things should quite drastically reduce POLREF alignment-time requirements, which currently take up quite a bit of time _for each sample_ (i.e. a very significant amount of beamtime across a cycle).

---

- [ ] Implement a detector-mapping DAE [_reducer_](https://isiscomputinggroup.github.io/ibex_bluesky_core/devices/dae.html#reducers) which pulls the entire spectrum-map for POLREF. See [`genie.get_spectrum_data`](https://github.com/ISISComputingGroup/genie/blob/723dbf863e12bf3ce95bb58f34cead23a09c9441/src/genie_python/genie_dae.py#L1119) for the underlying mechanism to do this.
- [ ] A `LiveDispatcher` which passes the data ~to [bumps](https://pypi.org/project/bumps/)~ (having talked to CK, we'll try a normal `lmfit` first, and only write a `bumps` fitter if we really need to:
  * Convert spectra number -> position, for 640 specified detector pixels, using the data  in `\\isis\inst$\NDXPOLREF\user\scripts\mantid\Calibrations\Cycle 18_3\OSMOND_Map_18_3_thetaScan_thVsCl_pixelsRemoved.nx5` in `workspace/axis1`
    * i.e. data in spectrum N is at an angle of 3.074 and spectrum N+639 is at angle 0.17392
    * The absolute numbers are actually not important to scientists - they are primarily interested in the _relative offsets_ between those numbers.
  * For each of the 640 spectra, integrate counts across all time channels to get an array of 640 counts&uncertainties
    * For efficiency in operation they will probably define just one time channel per pixel, but for diagnostics/comissioning/fault-finding they also want the option to be able to run this will full time-channel resolution too
[...]

PR #118, opened 2025-02-26
POLREF detector-mapping alignment

Tom-Willemsen:
Closes https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/103
Closes https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/113


Tom-Willemsen (commented):
Need to insert logging in various places

doc/index.rst
```diff
@@ -63,6 +63,13 @@ Reference documentation
 
    plan_stubs/*
 
+.. toctree::
+   :maxdepth: 2
+   :caption: Plans (Reflectometry)
+   :glob:
+
+   plans/reflectometry/*
```
rerpha, line 71: if #111 gets merged before it has the refl-specific plans just in plans.md, we should decide on a structure. 

maybe plans should be the toctree with a separate _file_ for technique specific stuff?

src/ibex_bluesky_core/plans/reflectometry/det_map_align.py
```diff
@@ -0,0 +1,241 @@
...
+
+    Args:
+        dae: The DAE to acquire from
+        angle_map: a numpy array, with the same shape as detectors,
+            describing the detector angle of each detector pixel
+
+    """
+    logger.info("Starting angle scan")
+    reducer = dae.reducer
+    _check_angle_map_shape(reducer, angle_map)
```
rerpha, line 126: is it worth adding a generic helper that checks the controller is a runperperiod controller and reducer is a PeriodSpecIntegralsReducer

we might want to use this in plans generally?

tests/test_plan_stubs.py
```diff
@@ -119,3 +127,26 @@ def plan():
...
+
+
+async def test_set_num_periods_works(RE, simpledae):
+    set_mock_value(simpledae.number_of_periods, 1)
+    RE(set_num_periods(simpledae, 5))
+    assert (await simpledae.number_of_periods.get_value()) == 5
+
+
+def test_set_num_periods_fails(RE, simpledae):
+    class UnequalInt(int):
```
rerpha, line 139: fixture for this? i could see it being helpful for other tests

src/ibex_bluesky_core/plans/reflectometry/det_map_align.py
```diff
@@ -0,0 +1,241 @@
...
+        ]
+    )
+    def _inner() -> Generator[Msg, None, None]:
+        yield from bp.count([dae])
+
+    yield from _inner()
+
+    logger.info("Finished angle scan, result = %s", angle_fit.result)
+
+    return cast(ModelResult | None, angle_fit.result)
```
rerpha, line 147:
```suggestion
    return angle_fit.result
```

src/ibex_bluesky_core/plans/reflectometry/det_map_align.py
```diff
@@ -0,0 +1,241 @@
+"""Reflectometry detector-mapping alignment plans."""
+
+import logging
+from collections.abc import Generator
+from typing import TypedDict, cast
```
rerpha, line 5:
```suggestion
from typing import TypedDict
```

rerpha (commented):
some initial comments

tests/test_plan_stubs.py
Tom-Willemsen: This whole thing has gone away since we no longer have a `set_num_periods` plan stub

src/ibex_bluesky_core/callbacks/file_logger.py
```diff
@@ -38,17 +38,26 @@
 class HumanReadableFileCallback(CallbackBase):
     """Outputs bluesky runs to human-readable output files in the specified directory path."""
 
-    def __init__(self, fields: list[str], *, output_dir: Path = DEFAULT_PATH) -> None:
+    def __init__(
+        self, fields: list[str], *, output_dir: Path = DEFAULT_PATH, postfix: str = ""
```
rerpha, line 42: slightly pedantic, but the example in the docs for this callback doesn't specify a postfix, could you add one pls

pyproject.toml
```diff
@@ -47,7 +47,9 @@ dependencies = [
   "lmfit",
   "scipy",
   "numpy",
+  "orjson",
   "scipp",
+  "typing-extensions",
```
rerpha, line 52: i don't know if it's worth a comment next to this saying why this is required - we'll have to update it every time we want to go up a version of python..?

doc/index.rst
Tom-Willemsen: I've made an initial pass at this, I think a more general look at all of the docs from our 3 big PRs will be good to do this when all 3 PRs are in.

pyproject.toml
rerpha: was thinking we'd need to change the comment depending on if the new version has the feature we're relying on typing-extensions for. 

Do we actually need it though - it looks like we're using it for TypeVar?

rerpha: yes we do for `default`

rerpha: ok - ignore me, sorry, i think im confusing things

Tom-Willemsen: We're using it for `TypeVar` that allows a `default`, yes. That's a python 3.13+ feature otherwise.

The reason I used it is so that we can still take arguments of type `SimpleDae` (which allows any reducer/waiter/controller), but _also_ explicitly specify which ones are wanted for plans where that makes sense

rerpha: makes sense - i was just wondering whether it's worth adding that as a comment next to this line, but maybe that'd be stupid - we don't just justify other dependencies

Tom-Willemsen: Not stupid - just added justifications for all of our main deps.

rerpha:
image

src/ibex_bluesky_core/callbacks/_utils.py
```diff
@@ -3,8 +3,7 @@
 from platform import node
 
 INSTRUMENT = node()
-
-OUTPUT_DIR_ENV_VAR = "IBEX_BLUESKY_OUTPUT_DIR"
+OUTPUT_DIR_ENV_VAR = "IBEX_BLUESKY_CORE_OUTPUT"
```
rerpha, line 6: as this is done at import-level i think it's not possible to change the value of this env var and get it picked up dynamically ie. just before running a plan

doc/plans/reflectometry/detector_mapping_alignment.md
```diff
@@ -0,0 +1,274 @@
...
+    )
+
+    prefix = get_pv_prefix()
+    dae = SimpleDae(
+        prefix=prefix,
+        controller=controller,
+        waiter=waiter,
+        reducer=reducer,
+    )
+
```
rerpha, line 159: mv set_num_periods here plz

doc/plans/plans.md
```diff
@@ -7,8 +7,7 @@ All of these plans can be used directly by the RunEngine (`RE()`) or within a wr
 So you can do this: 
 
 ```python
-
-    >>> result = RE(refl_scan("S1VG", 1, 10, 21, model=Gaussian().fit() frames=500, det=100, mon=3, pixel_range=6, periods=True, save_run=False))
+result = RE(refl_scan("S1VG", 1, 10, 21, model=Gaussian().fit() frames=500, det=100, mon=3, pixel_range=6, periods=True, save_run=False))
```
rerpha, line 10: i guess it would make more sense to call motor_scan here now we've moved refl_scan out to another docs page - minor thing really

src/ibex_bluesky_core/callbacks/reflectometry/_det_map.py
```diff
@@ -0,0 +1,217 @@
...
+        self.y_out_name: str = y_out_name
+
+        self._descriptor_uid: str | None = None
+
+    def descriptor(self, doc: EventDescriptor) -> None:
+        """Process a descriptor."""
+        self._descriptor_uid = doc["uid"]
+        return super().descriptor(doc)
+
+    def event(self, doc: Event, **kwargs: dict[str, Any]) -> None:  # pyright: ignore
```
rerpha, line 112: return doc plz

rerpha: and remove pyright:ignore

src/ibex_bluesky_core/plan_stubs/__init__.py
```diff
@@ -73,5 +76,4 @@ def call_qt_aware(
     if not getattr(func, "__module__", "").startswith("matplotlib"):
         raise ValueError("Only matplotlib functions should be passed to call_qt_aware")
 
-    yield from bps.clear_checkpoint()
```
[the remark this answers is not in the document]
rerpha: yes, but #137

rerpha (approved):

Issue #107, opened 2025-02-10
[Investigate/PoC] DAE reducer which uses mantid via FIA API


Tom-Willemsen:
We've been using scipp for simple data reductions (e.g. `I/I_0`, etc), which is very performant as it can run in-process and has been quite easy to work with for those simple reductions.

However, for more complex data reductions, we likely want to be able to call out to a hosted mantid service too, so that scientists can use their existing full mantid reduction pipelines (we don't want to have to duplicate entire existing mantid workflows in scipp just to use them as part of bluesky scans)

---

There is a [`mantid-as-a-service` API](https://reduce.isis.cclrc.ac.uk/fia) available (*), which runs on central compute resource (i.e. not the NDX), with on-call support from the relevant team. Investigate whether or not this can plausibly be used for "complex" DAE data reductions/feedback during a bluesky scan. "complex" here just means that we don't want to duplicate the reduction in scipp - not necessarily related to _computational_ complexity - anything that runs synchronously with a scan will need to be quite fast anyway.

(*) It's actually a lot more general than just mantid - it can run any container, of which mantid happens to be one - but mantid is probably the most interesting bit for us right now.

---

Example questions to answer:
- Is is simple/maintainable enough to form the requests to the above API? 
  * e.g. do we need to ask relevant team to add extra API endpoints for us?
  * I think the minimum we need is something along the lines of
    * Submit job API `Script -> JobID`
    * Get status API `JobID -> Status [e.g. done/failed/in progress]` which we can poll regularly
    * Get result API `JobID -> str`
  - Current model is just to specify a container image + script to run, as a string. Seems like this is quite flexible at first glance, but may be a bit inefficient at times.
- How does the authN/authZ work?
  * uows functional account has been added as having permissions, for the moment. May want to think harder about this long term.
- Is it performant enough to realistically be used during a scan?
  * "performant enough" will depend strongly on the instrument
  * in general anything longer than a scan points' duration will be too slow as reduction jobs would "back up" during a scan and cause a long delay at scan end
  * much more than a few seconds won't be able to run _synchronously_ with the scan in most cases - i.e. couldn't be used in adaptive scans etc - but might still be viable to run asynchronously while the next point collects.
- Is it - and all the intervening infrastructure that this workflow would depend on - stable enough that we could _actually_ rely on it in practice?

This is very much an early "proof-of-concept", to show whether or not we _could_ do this as/when concrete requirements arise.

### Acceptance criteria
[...]

PR #122, opened 2025-02-28
PoC mantid reducer

Tom-Willemsen:
Proof-of-concept for mantid reduction as part of a bluesky plan using FIA api

Closes https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/107

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,12 +41,14 @@ classifiers = [
 ]
 
 dependencies = [
+  "aiohttp",
   "bluesky",
   "ophyd-async[ca] >= 0.8.0",
   "matplotlib",
   "lmfit",
   "scipy",
   "numpy",
+  "orjson",
   "scipp",
   "tzdata",
   "scippneutron",
--- a/src/ibex_bluesky_core/devices/dae/dae.py
+++ b/src/ibex_bluesky_core/devices/dae/dae.py
@@ -1,8 +1,8 @@
 """ophyd-async devices for communicating with the ISIS data acquisition electronics."""
 
 from numpy import int32
-from ophyd_async.core import Array1D, SignalR, SignalRW, StandardReadable, StrictEnum
-from ophyd_async.epics.core import epics_signal_r, epics_signal_rw
+from ophyd_async.core import Array1D, SignalR, SignalRW, SignalW, StandardReadable, StrictEnum
+from ophyd_async.epics.core import epics_signal_r, epics_signal_rw, epics_signal_w
 
 from ibex_bluesky_core.devices import isis_epics_signal_rw
 from ibex_bluesky_core.devices.dae.dae_controls import DaeControls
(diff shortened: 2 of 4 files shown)


Tom-Willemsen:
Initial thoughts:
- Currently we have to do a binary search to find last-submitted job, and _hope_ that's our job. Not great.
- Latency is about 10-15 seconds, plus however long the actual mantid reduction script takes. Quite high variability, sometimes it's just a few seconds or sometimes up to 30s. Guess this might just be a function of load on the cluster.
- Currently passing various data around as json/strings. This isn't very efficient (and FIA outright rejects medium-large spectra-maps if we just do the dumb "embed them in the script" approach. Need to do something cleverer, likely with input files. Unclear if FIA api _currently_ has infrastructure to help us do this (beyond just saving out a full nexus, waiting for it to show up on archive, and getting mantid to read that)
- Various bits of FIA api are only really documented in their source code right now - e.g. exactly what you have to print to get "results" to show up in output.
- I think this probably _is_ viable, but only for use cases with quite long runs. 15s latency is way too much for e.g. reflectometry alignment scans.

Tom-Willemsen:
Some answers from my mantid/FIA contact...

> Latency is about 10-15 seconds, plus however long the actual mantid reduction script takes. Quite high variability, sometimes it's just a few seconds or sometimes up to 30s. Guess this might just be a function of load on the cluster.

This is expected and not easy to reduce.

> Currently passing various data around as json/strings. This isn't very efficient (and FIA outright rejects medium-large spectra-maps if we just do the dumb "embed them in the script" approach. Need to do something cleverer, likely with input files. Unclear if FIA api currently has infrastructure to help us do this (beyond just saving out a full nexus, waiting for it to show up on archive, and getting mantid to read that - but that'll be very slow)

The reduction process will need to map the instrument data area soon anyway, so we can just dump arrays there rather than embedding them.

> Various bits of FIA api are only really documented in their source code right now - e.g. exactly what you have to print to get "results" to show up in output.

Yes the documentation is basically just the source code at present and what we're doing is "correct" in the sense that it works.

> Right now we're abusing output_files to return answers that aren't filenames. That's probably bad.

Some suggestion of using `outputs` in future rather than `output_files` but fundamentally what we're doing isn't "wrong".

> I think this probably is viable, but only for use cases with quite long runs. 15s latency is way too much for e.g. reflectometry alignment scans.

Yes this is true, mantid reductions won't be viable for short runs, this is unlikely to change in short term. Long term there _might_ be other solutions.

> FIA will briefly give bad http responses when it gets redeployed, we'll need to put in some careful retry logic to get around this. Have asked my contact what kinds of retry delays/timeouts we might need.

Should be up again within ~seconds. Suggest retry loop retries 20 times at 5s intervals or something before erroring if API is dead.

Tom-Willemsen:
This approach is not viable to be used in real-life; site network and cloud infrastructure are not reliable enough at present.

Can be revisited if MNeuData defines more reliable places to run the analysis - e.g. compute resource in instrument cabins or halls.

PR #124, opened 2025-03-06
codeql

Tom-Willemsen:
Maybe try this here first, and if it works well, then promote to `reusable-workflows` so that all python repos get it for free.


rerpha (approved):

rerpha:
i see no harm in adding this to reusable workflows

PR #133, opened 2025-03-13
Polref Full Auto Alignment Plans

jackbdoughty:
This PR implements functionality for reflectometers to be able to auto align their beamlines.

 It introduces the following:
- The optimise_axis_against_intensity method. This is used for optimising an AlignmentParam against the beam. It will run atleast one scan (depending on the amount of scan ranges passed), attempt to find where the optimal motor position is (given which fitting parameter to aim to optimise), move to the position (if found to be sensible using a checking function). It has functionality for providing a function to check if the optimised valuue is sensible and if not then has option to call a user provided plan.

### DIFF
--- a/.github/workflows/Lint-and-test.yml
+++ b/.github/workflows/Lint-and-test.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ "ubuntu-latest", "windows-latest" ]
-        version: ['3.11', '3.12']
+        version: ['3.10', '3.11', '3.12']
       fail-fast: false
     steps:
       - uses: actions/checkout@v4


jackbdoughty:
TODO:
- Figure out why NotRequired doesn't work on linux tests
- Add more parameters and parameter tests to polref plan
- Figure out why docs don't want to build, then make some docs
- Figure out why complaining about MYPVPREFIX in tests, should not need this

Tom-Willemsen (commented):
Slightly struggling to see what this will actually look like at the "top level" - did you have an example top-level for POLREF showing how all this stuff actually gets called?

doc/plans/refl_plans.md
```diff
@@ -0,0 +1,181 @@
...
+
+my_device = AlignmentParam(
+    name="my_device",
+    rel_scan_ranges=[2.0],
+    fit_method=Gaussian().fit(),
+    fit_param="x0"
+    )
+```
+
+`name` - Should be the same as the corresponding reflectometry parameter.
```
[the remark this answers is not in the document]
jackbdoughty: good point, maybe worth adding functionality to support namedmovable aswell as reflparamater

src/ibex_bluesky_core/plans/reflectometry/autoalign_utils.py
```diff
@@ -0,0 +1,357 @@
...
+        save_run (bool): Should runs be saved. Defaults to True.
+        files_output_dir (Path): Where to save any outputted files. Defaults to
+            C:/instrument/var/logs/bluesky/output_files.
+        problem_found_plan (Callable[[], Generator[Msg, None, None]] | None):
+            Either a plan or standard function, called if optimised value is not found to be
+            sensible.
+        pre_align_plan (Callable[[], Generator[Msg, None, None]] | None):
+            Either a plan or standard function, called before all scans.
+        post_align_plan (Callable[[], Generator[Msg, None, None]] | None):
+            Either a plan or standard function, called after all scans.
```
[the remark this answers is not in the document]
Tom-Willemsen: I think I was more commenting on

> Either a plan or standard function

Where we now said it would just always be a plan

doc/plans/refl_plans.md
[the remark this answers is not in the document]
rerpha: reflparameter should be a namedMovable?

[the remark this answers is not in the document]
rerpha: (so can just used NamedMovable as the type)

src/ibex_bluesky_core/plans/reflectometry/autoalign_utils.py
```diff
@@ -0,0 +1,357 @@
...
+            *chain(
+                *[
+                    (p.get_movable(), self.pre_align_param_positions[p.name])
+                    for p in alignment_params
+                ]
+            )  # type: ignore
+        )
+
+
+def _add_fields(
```
rerpha, line 98: this part is sort of duplicated by https://github.com/IsisComputingGroup/ibex_bluesky_core/blob/c85e1a91060c5af252e5d5aec4b2aa2c73cb8565/src/ibex_bluesky_core/plans/__init__.py#L70 but maybe we should pull the fields bit out so you can just pass in more stuff (ie `alignment_param_name`) - up to you really

tests/plans/test_reflectometry.py
```diff
@@ -60,3 +79,467 @@ def test_refl_adaptive_scan_creates_refl_param_device_and_dae(RE):
         assert isinstance(scan.call_args[1]["dae"], SimpleDae)
         assert isinstance(scan.call_args[1]["block"], ReflParameter)
         assert scan.call_args[1]["block"].name == param_name
+
+
+# Auto-Alignment Utils
+
+
+@pytest.fixture
+async def dae():
```
rerpha, line 88: https://github.com/ISISComputingGroup/ibex_bluesky_core/blob/main/tests/conftest.py#L23 could use this instead

src/ibex_bluesky_core/plans/reflectometry/autoalign_utils.py
```diff
@@ -0,0 +1,357 @@
...
+    the process for a smaller range and higher granularity.
+
+    Args:
+        dae (SimpleDae | BlockR[float]): A readable signal that represents beam intensity.
+        alignment_param (AlignmentParam): The alignment parameter to be scanned over and optimised.
+
+    Keyword Args:
+        num_points (int): The number of points across the scan. Defaults to 10.
+        fields (list[str]): Fields to measure and document in outputted files.
+        periods (bool): Are periods being used. Defaults to True.
```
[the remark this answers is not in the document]
rerpha:
```suggestion
        periods (bool): Are DAE software periods being used. Defaults to True.
```

doc/plans/refl_plans.md
```diff
@@ -0,0 +1,181 @@
...
+yield from my_device.pre_align_params(parameters)
+# This will move my_device to 1.0 and my_device2 to 2.0 simultaneously
+
+```
+
+Note that there is also a `get_movable` method which returns a `AlignmentParam`'s underlying `ReflParameter `object.
+
+## Full-Auto Alignment Plan
+
+The [`optimise_axis_against_intensity`](ibex_bluesky_core.plans.reflectometry.autoalign_utils.optimise_axis_against_intensity) plan by default scans over a motor against beam intensity, and given which fitting parameter chosen to be optimised and the fitting method, it will find aim to find an optimum value. At this stage it will check if the value is 'sensible'- there are some default checks but the user is encouraged to provide their own checks per `AlignmentParam` as described in the next section. If found to be sensible, the motor will be moved to this value, and the motor position redefined as 0.
```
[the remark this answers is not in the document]
jackbdoughty: I think given that at the moment I'm only reading/writing to ReflParameters it'll be on the reflectometry level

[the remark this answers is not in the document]
Tom-Willemsen: Is that what their existing auto-align scripts do?

jackbdoughty:
> Slightly struggling to see what this will actually look like at the "top level" - did you have an example top-level for POLREF showing how all this stuff actually gets called?

Here is my top level POLREF script as of right now. I'll expect to add some more parameters and checking functions:

```python
"""Auto-alignment scripts for POLREF reflectometer."""

from collections.abc import Generator
from datetime import datetime
from math import isclose
from pathlib import Path
from winsound import Beep

import matplotlib.pyplot as plt
from bluesky import plan_stubs as bps
from bluesky.utils import Msg
from lmfit.model import ModelResult
from ophyd_async.plan_stubs import ensure_connected

from ibex_bluesky_core.callbacks.fitting.fitting_utils import SlitScan
from ibex_bluesky_core.devices.simpledae import monitor_normalising_dae
from ibex_bluesky_core.plan_stubs import call_sync
from ibex_bluesky_core.plans.reflectometry import autoalign_utils, centred_pixel
from ibex_bluesky_core.utils import get_pv_prefix

PREFIX = get_pv_prefix()
COUNT = 20
FRAMES = 500
READABLE_FILE_OUTPUT_DIR = (
    Path("C:\\") / "instrument" / "var" / "logs" / "bluesky" / "output_files" / ""
)
DEFAULT_DET = 280
DEFAULT_MON = 2
PIXEL_RANGE = 5
PERIODS = True
SAVE_RUN = True
NUM_POINTS = 21


def s1vg_checks(result: ModelResult, alignment_param_value: float) -> bool:
    """Check for optimised S1VG value."""
    rsquared_confidence = 0.9
    expected_param_value = 1.0
    expected_param_value_tol = 0.1
    max_peak_factor = 5.0

    if result.rsquared < rsquared_confidence:
        return True

    # For S1VG, provide a value you would expect for it,
    # assert if its not close by a provided factor
    if not isclose(expected_param_value, alignment_param_value, abs_tol=expected_param_value_tol):
        return True

    # Is the peak above the background by some factor (optional because param may not be for
    # a peak, or background may not be a parameter in the model).
    if result.values["background"] / result.model.func(alignment_param_value) <= max_peak_factor:
        return True

    return False


parameters = [
    autoalign_utils.AlignmentParam(
        name="S1VG",
        rel_scan_ranges=[0.3, 0.05],
        fit_method=SlitScan.fit(),
        fit_param="inflection0",
        is_good_fit=s1vg_checks,
        pre_align_param_positions={
            "THETA": 0,
            "PHI": 0,
            "S1VG": -0.1,  # 0.2?
            "S2VG": 10,
            "S3N": 5,
            "S3S": -5,
            "S1HG": 40,
            "S2HG": 30,
            "S3E": 15,
            "S3W": 15,
        },
    ),
]


def full_autoalign_plan() -> Generator[Msg, None, None]:
    """Full autoalign plan for POLREF."""
    det_pixels = centred_pixel(DEFAULT_DET, PIXEL_RANGE)
    dae = monitor_normalising_dae(
        det_pixels=det_pixels,
        frames=FRAMES,
        periods=PERIODS,
        save_run=SAVE_RUN,
        monitor=DEFAULT_MON,
    )

    yield from bps.mv(dae.number_of_periods, COUNT if PERIODS else 1)  # type: ignore

    axes_sig = [parameters[i].get_movable() for i in range(len(parameters))]
    yield from ensure_connected(*axes_sig, dae)  # type: ignore

    print("Starting auto-alignment...")

    for i in range(len(parameters)):
        print(f"Next alignment parameter is {parameters[i].name}")
        Beep(2500, 200)

        yield from autoalign_utils.optimise_axis_against_intensity(
            dae=dae,
            alignment_param=parameters[i],
            fields=[],
            periods=PERIODS,
            save_run=SAVE_RUN,
            num_points=NUM_POINTS,
            files_output_dir=READABLE_FILE_OUTPUT_DIR,
            problem_found_plan=lambda: call_sync(Beep, 1000, 500),
            pre_align_plan=lambda i=i: parameters[i].pre_align_params(parameters),
            post_align_plan=lambda i=i: call_sync(
                plt.savefig,
                fname=f"{READABLE_FILE_OUTPUT_DIR}{parameters[i].name}_{datetime.now().strftime('%H%M%S')}",
            ),
        )

    print("Auto alignment complete.")

```

src/ibex_bluesky_core/plans/reflectometry/autoalign_utils.py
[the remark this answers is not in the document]
jackbdoughty: Oh yep sorry, didn't notice the description underneath the type

doc/plans/refl_plans.md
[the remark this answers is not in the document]
jackbdoughty: Would there be any downsides to just making AlignmentParam inherit from NamedMovable[float]? and then I don't need a method to retrieve the underlying signal

doc/plans/refl_plans.md
```diff
@@ -0,0 +1,113 @@
...
+        frames=FRAMES,
+        periods=PERIODS,
+        save_run=SAVE_RUN,
+        monitor=DEFAULT_MON,
+    )
+
+    s1vg = ReflParameter(prefix=PREFIX, name="S1VG", changing_timeout_s=60)
+    # Interfaces with the reflectometry server
+
+    yield from bps.mv(dae.number_of_periods, count)
```
rerpha, line 29: don't think you need to do this here - the `scan` plan does it for you if using periods

src/ibex_bluesky_core/plans/reflectometry/autoalign_utils.py
```diff
@@ -0,0 +1,232 @@
...
+        init_mot_pos (float): The initial motor position before scanning.
+        rel_scan_range (float): The current relative scan range.
+        is_good_fit (Callable[[ModelResult, float], bool] | None): User provided checks on the
+            optimised value, must return True if result is sensible.
+
+    Returns:
+        True if value is sensible. False otherwise.
+
+    """
+    if init_mot_pos + rel_scan_range < alignment_param_value:
```
rerpha, line 42: may want to raise a `ValueError` here with a useful message that is caught outside - otherwise a user might be "in the dark" as to what has failed

rerpha: maybe make a new exception for this

doc/plans/refl_plans.md
```diff
@@ -0,0 +1,113 @@
...
+
+```
+
+Note that {py:obj}`ibex_bluesky_core.plans.reflectometry.autoalign_utils.optimise_axis_against_intensity` also has parameters:
+
+`periods` - Whether to use periods.
+
+`save_run` - Whether to save runs.
+
+`problem_found_plan` - A callback plan for if the optimised value is not found to be sensible.
```
rerpha, line 102: maybe don't need these here, they are covered by the API reference

src/ibex_bluesky_core/plans/reflectometry/autoalign_utils.py
```diff
@@ -0,0 +1,232 @@
...
+    alignment_param: NamedMovable[float],
+    fit_method: FitMethod,
+    fit_param: str,
+    rel_scan_ranges: list[float],
+    **kwargs: Unpack[OptimiseAxisParams],
+) -> Generator[Msg, None, ISISCallbacks | None]:
+    """Optimise a motor/device to the intensity of the beam.
+
+    Scan between two symmetrical points relative to alignment_param's current motor position,
+    to find where the optimal point is depending on alignment_param.fit_param, move to it and
```
rerpha, line 174:
```suggestion
    to find where the optimal point is depending on fit_param, move to it and
```

src/ibex_bluesky_core/plans/reflectometry/autoalign_utils.py
```diff
@@ -0,0 +1,232 @@
...
+    """Type hints for optimise_axis_against_intensity."""
+
+    num_points: NotRequired[int]
+    periods: NotRequired[bool]
+    save_run: NotRequired[bool]
+    problem_found_plan: NotRequired[Callable[[], Generator[Msg, None, None]]]
+    is_good_fit: NotRequired[Callable[[ModelResult, float], bool] | None]
+
+
+def optimise_axis_against_intensity(  # noqa: D417
```
rerpha, line 163: probably need to say in the docs that this moves the movable to the optimum position, so you _can_ redefine it in your top level plan if you want to

doc/plans/refl_plans.md
```diff
@@ -0,0 +1,112 @@
...
+
+To define a `problem_found_plan` you must use the following function signature:
+
+```python
+
```
rerpha, line 107: we should have an example of how to use one of these

src/ibex_bluesky_core/plans/reflectometry/autoalign_utils.py
```diff
@@ -0,0 +1,232 @@
...
+    dae: SimpleDae,
+    alignment_param: NamedMovable[float],
+    fit_method: FitMethod,
+    fit_param: str,
+    rel_scan_ranges: list[float],
+    **kwargs: Unpack[OptimiseAxisParams],
```
rerpha, line 169: might be better to use keyword-only args here rrather than OptimiseAxisParams

src/ibex_bluesky_core/plans/reflectometry/autoalign_utils.py
```diff
@@ -0,0 +1,232 @@
...
+            choice = yield from call_sync(
+                input,
+                """Type '1' if you would like to re-scan or type '2' to re-zero"""
+                f""" at {alignment_param_value} and keep going.""",
+            )
+
+            if choice not in {"1", "2"}:
+                choice = yield from inp()
+
+            return choice
```
rerpha, line 141: it might make sense to pull this out as a plan stub and make it generic - can see it being useful elsewhere

Tom-Willemsen:
Now ruff has hopefully sorted itself out do we actually still need to drop 3.10?

Tom-Willemsen (approved):

Tom-Willemsen:
Let's get this through for the moment... inevitably it'll need further changes when we test but that's ok

PR #139, opened 2025-03-19
add plan stub to set a motor's current position (redefine)

rerpha:
we would need to use ophyd async latest as my changes to expose the set/use switch aren't in a release yet.


pyproject.toml
```diff
@@ -42,7 +42,7 @@ classifiers = [
 
 dependencies = [
   "bluesky",  # Bluesky framework
-  "ophyd-async[ca] >= 0.8.0",  # Device abstraction
+  "ophyd-async[ca,pva] @ git+https://github.com/bluesky/ophyd-async@bf6b71f045bf297e2bbbfd3ea4de7f368643a948",  # Device abstraction
```
rerpha, line 45: Comment here for pin?

Tom-Willemsen (approved):

esmith1729 (approved):

PR #140, opened 2025-03-20
add callback for saving plots to PNGs

### LINKED ISSUE #132: callback for saving a plot to PNG on an end document
As a user I would like to be able to output my finished plots to a PNG file as they are useful in logbooks 

This should be a separate callback which is toggle-able from `ISISCallbacks` but enabled by default(?)

`pyplot.savefig()` is all we need to do here on a `plt` which is passed into the callback.

rerpha:
Closes #132 
i also refactored out format_date() as we were doing it in several places. technically this changes the human readable file format but i don't think that matters yet


src/ibex_bluesky_core/callbacks/plotting.py
```diff
@@ -74,3 +79,24 @@ def start(self, doc: RunStart) -> None:
         """Process an start document (delegate to superclass, then show the plot)."""
         super().start(doc)
         show_plot()
+
+
+class PlotPNGSaver(CallbackBase):
```
Tom-Willemsen, line 84: You might need to make this derive from `QtAwareCallback` if you want to avoid race conditions where you try to save the plot before it's been drawn (since the actual plotting is done in a QtAware, and therefore potentially on other threads, meaning it hasn't necessarily finished by the time you get here

rerpha:
ruff will be fixed by https://github.com/astral-sh/ruff/issues/16874 when it's released

src/ibex_bluesky_core/callbacks/plotting.py
```diff
@@ -74,3 +86,49 @@ def start(self, doc: RunStart) -> None:
...
+            x: The name of the signal for x.
+            y: The name of the signal for y.
+            ax: The subplot to save to a file.
+            postfix: The file postfix.
+            output_dir: The output directory for PNGs.
+
+        """
+        super().__init__()
+        if ax is None:
+            raise ValueError("ax is None")
```
Tom-Willemsen, line 114: Why do we type-hint `ax: Axes | None,` if we just throw an error if it is `None`?

src/ibex_bluesky_core/callbacks/plotting.py
```diff
@@ -74,3 +86,49 @@ def start(self, doc: RunStart) -> None:
...
+        self.output_dir = Path(output_dir or get_default_output_path())
+
+    def stop(self, doc: RunStop) -> None:
+        """Write the current plot to a PNG file.
+
+        Args:
+            doc: The stop document.
+
+        """
+        rb_num = _get_rb_num(doc)
```
Tom-Willemsen, line 128: Does this actually work - there is no rb_num in a stop document?

It's in the _start_ document, so you'll have to handle a start doc too and stash the RB number from it so that you can use it in this cb.

tests/callbacks/test_plotting_callback.py
```diff
@@ -77,6 +81,35 @@ def test_errorbars_created_if_yerr_is_given():
...
+                "x": 1,
+                "y": 2,
+            },  # pyright: ignore [reportArgumentType]
+            "rb_number": 1234,  # pyright: ignore [reportArgumentType]
+            "time": 123456789,
+            "uid": "1",
+            "exit_status": "success",
+            "run_start": "",
+        }
+    )
```
Tom-Willemsen, line 101: `pyright` is correct here, neither `data` nor `rb_number` belong in a stop document.

rerpha:
put in `PNGS/`

PR #145, opened 2025-03-25
remove duplicate fields

rerpha:
we specify them as `y` and `yerr` anyway, so they end up in a human-readable file header twice. No need to explicitly add them.


jackbdoughty (approved):
Can't see any problem with this

PR #147, opened 2025-03-25
add device for manager mode

### LINKED ISSUE #146: reflectometry redefine: requires manager mode
I can't redefine a manager mode and ophyd async fails saying it doesnt have write exception if manager mode is not disabled - this is refl server functionality. we should probably make our redefine_refl_parameter have some sort of finalise wrapper which sets it then unsets it afterwards.

rerpha:
closes #146

### DIFF
--- a/src/ibex_bluesky_core/devices/reflectometry/__init__.py
+++ b/src/ibex_bluesky_core/devices/reflectometry/__init__.py
@@ -13,6 +13,7 @@ from ophyd_async.core import (
 )
 from ophyd_async.epics.core import epics_signal_r, epics_signal_w
 
+from ibex_bluesky_core.devices.manager_mode import ManagerModeChoices, manager_mode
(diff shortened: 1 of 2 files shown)


src/ibex_bluesky_core/devices/reflectometry/__init__.py
```diff
@@ -76,6 +77,7 @@ def __init__(self, prefix: str, name: str, changed_timeout_s: float = 5.0) -> No
         """
         self.changed: SignalR[bool] = epics_signal_r(bool, f"{prefix}DEFINE_POS_CHANGED")
         self.define_pos_sp = epics_signal_w(float, f"{prefix}DEFINE_POS:SP")
+        self.manager_mode = manager_mode()
```
rerpha, line 80: this is wrong

PR #177, opened 2025-04-02
allow 'exit now' as an option for optimise_axis_over_range()

### LINKED ISSUE #158: Optimize plan: add "exit now" option

rerpha:
Closes #158


src/ibex_bluesky_core/plans/reflectometry/_autoalign.py
```diff
@@ -146,12 +146,16 @@ def _optimise_axis_over_range(  # noqa: PLR0913 PLR0917
...
+            prompt=f"Type '1' if you would like to re-scan, '2' to "
+            f"move {alignment_param.name} to {alignment_param_value} and keep going, "
+            f"or '3' to exit now.",
+            choices=["1", "2", "3"],
         )
         if choice == "1":
             return icc, False
+        elif choice == "3":
+            _print_and_log("Plan terminated.")
+            raise GeneratorExit()
```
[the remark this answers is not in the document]
rerpha: that works really nicely. will implement

src/ibex_bluesky_core/plans/reflectometry/_autoalign.py
```diff
@@ -146,12 +146,16 @@ def _optimise_axis_over_range(  # noqa: PLR0913 PLR0917
...
+            prompt=f"Type '1' if you would like to re-scan, '2' to "
+            f"move {alignment_param.name} to {alignment_param_value} and keep going, "
+            f"or '3' to pause now.",
+            choices=["1", "2", "3"],
         )
         if choice == "1":
             return icc, False
+        elif choice == "3":
+            _print_and_log("Plan paused.")
+            yield from bps.pause()
```
Tom-Willemsen, line 158: So after a `RE.resume()` we move to the previously-aligned position (as if we'd typed `2`) - not sure that's right? I think it's more likely that we want to `return icc, False` here to re-scan after someone has fixed whatever needed fixing?

I could also see a case for asking the question again...

rerpha: true - i think exiting is the easiest thing to do here, a scan shouldn't really take that long to re-do after some sort of manual adjustment?

Tom-Willemsen: >  scan shouldn't really take that long to re-do after some sort of manual adjustment?

So that sounds like just `return icc, False` after the bps.pause then I think.

You might need a checkpoint just before the pause (to avoid replaying unwanted messages).

rerpha: Yes - just added one of those funnily

Tom-Willemsen (approved):

Issue #153, opened 2025-03-27
Unlive plot


Tom-Willemsen:
The angle map scan uses a `LivePlot` but it's not really meaninful to `Live` plot this as the events are all already there in memory.

Make it do a plot that only updates once at the end

PR #187, opened 2025-04-24
Optometrise

Tom-Willemsen:
Closes https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/153

- Adds an option to `LivePlot` to only plot at end
- Make refl detector-mapping stuff only fit at end
- Make sure we're not using [Schlemiel's algorithm](https://en.wikichip.org/wiki/schlemiel_the_painter%27s_algorithm) for error bar rendering
- Some Qt nonsense, because every PR needs some Qt nonsense

For a test case on my machine with ~1280 events, this reduces time spent on fitting/plotting callbacks from ~30s to ~1s

### DIFF
--- a/src/ibex_bluesky_core/callbacks/__init__.py
+++ b/src/ibex_bluesky_core/callbacks/__init__.py
@@ -68,7 +68,7 @@ class ISISCallbacks:
         live_fit_logger_postfix: str = "",
         human_readable_file_postfix: str = "",
         live_fit_update_every: int | None = 1,
-        live_plot_show_every_event: bool = True,
+        live_plot_update_on_every_event: bool = True,
     ) -> None:
         """A collection of ISIS standard callbacks for use within plans.
 
@@ -128,7 +128,7 @@ class ISISCallbacks:
             live_fit_logger_postfix: the postfix to add to live fit logger.
             human_readable_file_postfix: optional postfix to add to human-readable file logger.
             live_fit_update_every: How often, in points, to recompute the fit. If None, do not compute until the end.
-            live_plot_show_every_event: whether to show the live plot on every event, or just at the end.
+            live_plot_update_on_every_event: whether to show the live plot on every event, or just at the end.
         """  # noqa
         self._subs = []
         self._peak_stats = None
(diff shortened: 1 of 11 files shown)


tests/test_utils.py
```diff
@@ -20,3 +20,9 @@ def test_cannot_get_pv_prefix():
 
 def test_centred_pixel():
     assert centred_pixel(50, 3) == [47, 48, 49, 50, 51, 52, 53]
+
+
+@pytest.mark.parametrize("mpl_backend", ["qt", "something_else"])
```
rerpha, line 25: should we test with `qt4Agg`, `qt5Agg`, `qtCairo` here rather than just `qt` to test that we match them all?

src/ibex_bluesky_core/callbacks/__init__.py
```diff
@@ -164,31 +169,41 @@ def _inner():
...
+                    def start(self, doc: RunStart) -> None:
+                        nonlocal fig, ax
+                        try:
+                            plt.close("all")
+                            fig, ax = plt.subplots()
+                        finally:
+                            done_event.set()
+
+                cb = _Cb()
+                cb("start", {"time": 0, "uid": ""})
```
rerpha, line 187: will this end up in our raw event log? I wonder if it's worth putting some metadata saying what it's doing so we don't get confused by starts with 0 time amongst "useful" events

src/ibex_bluesky_core/callbacks/__init__.py
```diff
@@ -66,6 +67,8 @@ def __init__(
         live_fit_logger_output_dir: str | PathLike[str] | None = None,
         live_fit_logger_postfix: str = "",
         human_readable_file_postfix: str = "",
+        live_fit_update_every: int | None = 1,
+        live_plot_show_every_event: bool = True,
```
rerpha, line 71: I don't really like this naming - `live_plot_show_every_event` sounds like "dont show every event, just some of them". 

would `live_plot_update_on_every_event` be clearer?

rerpha (commented):
some initial comments

src/ibex_bluesky_core/callbacks/__init__.py
Tom-Willemsen: no

rerpha (approved):

PR #190, opened 2025-04-29
Polref mods

Tom-Willemsen:
To make sure we don't lose these


src/ibex_bluesky_core/callbacks/_plotting.py
```diff
@@ -98,3 +109,52 @@ def stop(self, doc: RunStop) -> None:
...
+        """Write the current plot to a PNG file.
+
+        Args:
+            doc: The stop document.
+
+        """
+        if self.filename is None:
+            raise ValueError("No filename specified for plot PNG")
+
+        self.ax.figure.savefig(self.filename, format="png")  # pyright: ignore [reportAttributeAccessIssue]
```
jackbdoughty, line 160: I feel like if we want to have a different directory for each rb number (as shown on line 146) then we should have the following before this line.
```python
if not os.path.isdir(self.filename):
    os.makedirs(self.filename)
```
because otherwise if the rb number folder doesn't already exist (which is likely right?) then we get a `FileNotFoundError: [Errno 2] No such file or directory:`

jackbdoughty (approved):

PR #192, opened 2025-05-01
use blockrw for motor scan instead of block_mot

rerpha:
This PR makes the motor_scan() and motor_adaptive_scan() consturct a BlockRw using the global moving flag. this change is required as if you try and use motor_scan() over something that is using axis.cmd it doesn't have all of the motor fields and then fails to connect to a load of fields that don't exist. this is what the SANS instruments want to do for most cases, so i think we should cater for it.

### DIFF
--- a/doc/plans/plans.md
+++ b/doc/plans/plans.md
@@ -67,7 +67,7 @@ Alternatively, we provide some very thin wrappers which construct the devices fo
 
 [`motor_adaptive_scan`](ibex_bluesky_core.plans.motor_scan)
 
-which wrap the above respectively. These take _names_ of blocks, rather than devices themselves, and construct a DAE and block device for you, using the global moving flag to determine if a motor has finished moving (in the same way as a `waitfor_move()`). This might be useful if you have a fairly standard DAE setup and just want to scan a block pointing at a motor such as a Sample Changer axis, but is obviously not as flexible or performant as the lower-level plans.
+which wrap the above respectively. These take _names_ of blocks, rather than devices themselves, and construct a DAE and block device for you, using the global moving flag to determine if a motor has finished moving (in the same way as a `waitfor_move()`). This might be useful if you have a fairly standard DAE setup and just want to scan a block pointing at a motor such as a Sample Changer axis, but is not as flexible or performant as the lower-level plans.
 
 for example if you just wanted to scan over a motor, wait for 400 frames, and perform a linear fit, you can just write this in the console: 
 
--- a/doc/plans/plans.md
+++ b/doc/plans/plans.md
@@ -67,7 +67,7 @@ Alternatively, we provide some very thin wrappers which construct the devices fo
 
 [`motor_adaptive_scan`](ibex_bluesky_core.plans.motor_scan)
 
-which wrap the above respectively. These take _names_ of blocks, rather than devices themselves, and construct a DAE and block device for you. This might be useful if you have a fairly standard DAE setup and just want to scan a block pointing at a motor such as a Sample Changer axis.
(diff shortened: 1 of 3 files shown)


doc/plans/plans.md
```diff
@@ -67,7 +67,7 @@ Alternatively, we provide some very thin wrappers which construct the devices fo
 
 [`motor_adaptive_scan`](ibex_bluesky_core.plans.motor_scan)
 
-which wrap the above respectively. These take _names_ of blocks, rather than devices themselves, and construct a DAE and block device for you. This might be useful if you have a fairly standard DAE setup and just want to scan a block pointing at a motor such as a Sample Changer axis.
+which wrap the above respectively. These take _names_ of blocks, rather than devices themselves, and construct a DAE and block device for you, using the global moving flag to determine if a motor has finished moving (in the same way as a `waitfor_move()`). This might be useful if you have a fairly standard DAE setup and just want to scan a block pointing at a motor such as a Sample Changer axis, but is obviously not as flexible or performant as the lower-level plans.
```
Tom-Willemsen, line 70:
```suggestion
which wrap the above respectively. These take _names_ of blocks, rather than devices themselves, and construct a DAE and block device for you, using the global moving flag to determine if a motor has finished moving (in the same way as a `waitfor_move()`). This might be useful if you have a fairly standard DAE setup and just want to scan a block pointing at a motor such as a Sample Changer axis, but is not as flexible or performant as the lower-level plans.
```

src/ibex_bluesky_core/plans/__init__.py
```diff
@@ -12,15 +12,15 @@
...
+from ibex_bluesky_core.devices.block import BlockMot, BlockRw, BlockWriteConfig
 from ibex_bluesky_core.devices.simpledae import monitor_normalising_dae
 from ibex_bluesky_core.fitting import FitMethod
 from ibex_bluesky_core.utils import NamedReadableAndMovable, centred_pixel, get_pv_prefix
 
 if TYPE_CHECKING:
     from ibex_bluesky_core.devices.simpledae import SimpleDae
 
-__all__ = ["adaptive_scan", "motor_adaptive_scan", "motor_scan", "polling_plan", "scan"]
+__all__ = ["BlockMot", "adaptive_scan", "motor_adaptive_scan", "motor_scan", "polling_plan", "scan"]
```
Tom-Willemsen, line 23:
```suggestion
__all__ = ["adaptive_scan", "motor_adaptive_scan", "motor_scan", "polling_plan", "scan"]
```

I don't think `BlockMot` should be exported here (people should be importing that from `devices.block`

src/ibex_bluesky_core/plans/__init__.py
```diff
@@ -192,7 +193,12 @@ def motor_scan(  # noqa: PLR0913
         an :obj:`ibex_bluesky_core.callbacks.ISISCallbacks` instance.
 
     """
-    block = BlockMot(prefix=get_pv_prefix(), block_name=block_name)
+    block = BlockRw(
+        float,
+        prefix=get_pv_prefix(),
+        block_name=block_name,
+        write_config=BlockWriteConfig(use_global_moving_flag=True),
+    )
```
Tom-Willemsen, line 201: Is this just `block_rw`?

src/ibex_bluesky_core/plans/__init__.py
rerpha: thanks pycharm

src/ibex_bluesky_core/plans/__init__.py
rerpha: yes

Tom-Willemsen (approved):

Issue #164, opened 2025-03-30
Plan wrappers for DAE setup changes with cleanup


Tom-Willemsen:
As a user I'd find it convenient to have a pre-written wrapper(s) for temporarily fiddling with the common DAE settings, and then putting them back to whatever they were before at the end of a plan.

The main ones would be:
- `with_num_periods(n)`
- `with_time_channels(tcbs)`
- `with_dae_tables(wiring, spectra, detector)`

We already have the functionality in raw signals to do all of these things - this ticket is to write plan wrappers / preprocessors which use that existing functionality but importantly also **remember what the settings were originally and put them back afterwards**.

POLREF already has a wrapper for the time-channels variant, so that one can be copied over from polref, docs/tests added, and then a similar pattern followed for the others.

Limit to just the commonly-changed settings listed above for this ticket - we can add more if/when we need them

## Acceptance Criteria
- [ ]  Wrappers described above are implemented in Bluesky, documented, and tests created

## Planning
08/05/2025 - 0:49:50

PR #213, opened 2025-06-24
Larmor spin echo plan

jackbdoughty:
spin echo plan for larmor src/ibex_bluesky_core/larmor_jack_plan.py
not intending to merge


jackbdoughty:
Coverage issue expected as no unit tests for LARMOR inst script/plan

rerpha:
Another thing - I don't think this sets the number of periods yet - it should probably be `num_points * 2`

jackbdoughty:
> Another thing - I don't think this sets the number of periods yet - it should probably be `num_points * 2` - I think we want to do it at this level. or by calling `with_num_periods` when #164 is merged?

i reckon waiting for #164 is a good idea for this

Issue #214, opened 2025-06-24
Archiving


Tom-Willemsen:
- As a systems/data administrator I would like scans data to be saved in a way that is appropriate for long-term, scientifically useful data
- As a scientist I would like scans data to be stored in a way that aligns with FAIR principles

---

This is a partial continuation of https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/24 , but focused on the archiving aspects.

**For avoidance of any doubt, we cannot tie these bluesky scan output files to the concept of a DAE run.**

On discussion with @ChrisM-S, an initial solution could be:
- Bluesky writes its files into `c:\data\<rb number>`. This avoids depending on network at the time when bluesky is writing the file.
- Bluesky marks files that it has finished with with a readonly flag
- Bluesky writes checksum files for all of the file(s) it has generated
- A cron job will copy/move the files from there into `\\isis\inst$\ndx<INST>\cycle_xxx\autoreduced`
  * To do: do we want any structure _under_ the `autoreduced` directory
- From there, the data will be replicated and archived using the existing mechanisms for autoreduced data
- File naming will stay as it is for the moment

### Acceptance criteria
- [x] Consult internally whether this is a workable solution
- [x] Discuss this proposed solution with instrument scientists for feedback
- [x] Write an ADR describing this decision, how/why we came to it, and what the tradeoffs are
- [x] Write tickets to do technical work required

PR #215, opened 2025-06-26
Draft ADR 7 (Document decisions relating to file archiving)

Tom-Willemsen:
### Description of work

An initial pass at an ADR describing file archiving decisions.

Once we are happy with this *internally*, I will merge it as a draft ADR and present it to scientists at a regular scans library catchup for feedback, and then once any feedback is incorporated it can become an active ADR.

### Ticket

https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/214

### Labels

*Add appropriate label(s) to this PR*

 - 'bluesky-documentation' - Document Changes

### Acceptance criteria

- [ ] ADR clearly specifies the problem that we are attempting to solve
- [ ] ADR clearly describes our proposed solution, including both positives and negatives, and *why* we came to that decision

### Documentation

See PR.


rerpha (dismissed):

doc/architectural_decisions/007-output-file-archiving.md
```diff
@@ -0,0 +1,271 @@
...
+
+Bluesky should generate checksums for each file it has finished writing, and insert those checksums into a windows
+alternative file stream, comparable to what is done for existing DAE data. These checksums
+can be used to check for data corruption as the files are moved to the archive, and later replicated between the
+archive servers.
+
+### Moving to the ISIS archive
+
+An automated cron task will look for read-only Bluesky output files, and their associated checksums, in `c:\data` at
+regular short intervals (for example, 1 minute), and will move them to:
```
FreddieAkeroyd, line 239: if we have one checksum file per data file then it can move checksum file, if we have a single file listing checksums it would be a bit more complicated

FreddieAkeroyd (dismissed):

doc/architectural_decisions/007-output-file-archiving.md
Tom-Willemsen: I guess if it is done as an alternative file stream then the checksum "automatically" gets moved along with the file...

FreddieAkeroyd: indeed, `linux` is the main reason i'd try and avoid them...

ChrisM-S: There are quite a few scenarios where the out of band checksum streams are useful currently (as they don't get included in the checksum calculation - even by accident).  They are also useful to ensure immutability between the instrument computer and the current archive - and I have on several occasions, run simple checks like these when the integrity of the file system is in question (e.g. disk errors on one archive server). Having a separate block of checksums in a different disk locality and directory structure might themselves be suspect (note: the check of the checksum/file stream is local and two-way - one validates the other).  A separate .zip of checksums might be the way to go, but it does provide it's own maintenance issues (however it does have a checksum itself and file changes can be made whilst updating checksums if necessary - so not unlike the second file stream).

rerpha (approved):

PR #219, opened 2025-07-01
remove set num periods in dae_scan.py

rerpha:
earlier in this test it sets up a RunPerPointController, so this is unnecessary


Tom-Willemsen (changes requested):
- Remove import of `bps`
- Remove comment that says "pyright doesn't understand"
- `¯\_(ツ)_/¯`

Tom-Willemsen (approved):

PR #224, opened 2025-07-02
Momentum fit

AWaseem18:
### Description of work done
- Added new muon momentum fitting method
- Created a new model and guess function for a muon momentum fit
- Create a unit test function for the model and guess functions
- Updated the docs with new documentation on the created model 

### Testing
- Spin up fake data to apply to the fit
- If you need the program for the fake data, speak to Jack Doughty
- Create a top level plan to run the scan
- Make sure Bob and Alice can be written and read from 
- Also try and test with noisy data to ensure it still converges well


src/ibex_bluesky_core/fitting.py
```diff
@@ -687,3 +688,58 @@ def guess(
...
+            x_slope = x[max_y:min_y]  # Gets all x values between the maximum and minimum y
+
+            if len(x_slope) != 0:
+                x0 = np.mean(x_slope)
+            else:
+                x0 = x[-1]  # Picked as it can't be 0
+
+            p = 1  # Expected value, not likely to change
+
+            w = (x[-1] - x[0]) / 9  # 9 is a rough estimate base on testing
```
rerpha, line 734: I think we should pull out magic numbers like this `9` here as constants - even if just as a variable above this statement.

doc/fitting/standard_fits.md
```diff
@@ -183,3 +183,22 @@ g(x) = \max(f(x), \text{background} - \text{height})
 ```{math}
 y = \min(g(x), \text{background})
 ```
+
+## Muon Momentum
+
+API Reference: ['MuonMomentum'](ibex_bluesky_core.fitting.MuonMomentum)
```
rerpha, line 189: This (backticks rather than quotes) makes it format like the others:

```suggestion
API Reference: [`MuonMomentum`](ibex_bluesky_core.fitting.MuonMomentum)
```

src/ibex_bluesky_core/fitting.py
```diff
@@ -687,3 +688,58 @@ def guess(
...
+    def guess(
+        cls, *args: int
+    ) -> Callable[[npt.NDArray[np.float64], npt.NDArray[np.float64]], dict[str, lmfit.Parameter]]:
+        """Momentum Scan Fit Guessing."""
+
+        def guess(
+            x: npt.NDArray[np.float64], y: npt.NDArray[np.float64]
+        ) -> dict[str, lmfit.Parameter]:
+            min_y = np.argmin(y)
+            max_y = np.argmax(y)
```
rerpha, line 720: I might be tempted to rename these to `index_min_y` and `index_max_y` so it's obvious they're array indices rather than the actual min x or y values

rerpha (commented):
Looks good - just have some initial comments from skimming the code

src/ibex_bluesky_core/fitting.py
```diff
@@ -687,3 +688,58 @@ def guess(
...
+
+            b = y[min_y]
+            r = y[max_y] - b
```
Tom-Willemsen, line 723: These are probably clearer as:
```suggestion
            b = np.min(y)
            r = np.max(y) - b
```

src/ibex_bluesky_core/fitting.py
```diff
@@ -687,3 +688,58 @@ def guess(
...
+
+            x_slope = x[max_y:min_y]  # Gets all x values between the maximum and minimum y
```
Tom-Willemsen, line 725: A scan isn't guaranteed to be contiguous left-to-right - a right-to-left scan will have x[0] > x[-1] (and an adaptive scan might have x and y arrays effectively "shuffled").

If you need to get contiguous, ascending x values and their corresponding y-values, use `np.argsort`.

tests/callbacks/fitting/test_fitting_methods.py
```diff
@@ -823,3 +824,145 @@ def test_guess_given_flat_data(self):
             outp = NegativeTrapezoid.guess()(x, y)
             # check that with flat data gradient guess is 0
             assert outp["gradient"] == pytest.approx(0.0, rel=1e-2)
+
+
+@pytest.fixture
+def values() -> dict[str, float]:
```
Tom-Willemsen, line 830: Not sure these fixtures belong at the top-level of this file (especially with relatively ambiguous names like `x` or `values`). Do they need to be fixtures, or could they be constants within the `TestMuonMomentum` class?

rerpha:
@jackbdoughty @AWaseem18 Do you have any code you both used to test this with? ie. a scan over a PV with that sort of shape? 
also, I think https://github.com/ISISComputingGroup/ibex_bluesky_core/pull/224#discussion_r2182609348 still needs to be addressed, i think you can get rid of the index vars entirely if you use `np.argsort`

PR #226, opened 2025-07-03
impoved erfc stretch guess

AWaseem18:
### Description of work

- Using a new formula to find stretch:
    - determines how much to stretch or compress a standard error function
    by comparing the x-range in your data to the x-range of a standard erf function
    over the same y-value interval
- updates docs & unit tests
- makes a couple other random changes e.g removing .value on lmfit parameters in unit tests & importing numpy polynomial as p


src/ibex_bluesky_core/fitting.py
```diff
@@ -464,13 +463,31 @@ def guess(
...
+            y95 = np.min(y) + 0.95 * dy
+
+            ind05 = np.argmin(np.abs(y - y05))
+            ind95 = np.argmin(np.abs(y - y95))
+
+            x05 = x[ind05]
+            x95 = x[ind95]
+
+            erfc_const = 3  # The plotted erfc function where the greatest change
+            # in y happens in the region -1.5 and 1.5
```
Tom-Willemsen, line 481: Please can we justify this constant mathematically?

If I follow naive logic with percentiles, this should be the difference between `erfinv(-0.9)` (aka -1.163...), and `erfinv(0.9)` (1.163), but that gives me 2.326 - not 3 - have you rounded that to 3? Or is there more maths I'm missing here?

Tom-Willemsen (commented):
Seems like there should be a unit test (which fails with the old implementation and passes with the new one)?

src/ibex_bluesky_core/fitting.py
```diff
@@ -736,16 +746,14 @@ def guess(
...
+            w = 1 / calculate_erf_stretch(x, y, erfc=True, pre_sorted=True)
 
             init_guess = {
                 "b": lmfit.Parameter("b", b),
                 "r": lmfit.Parameter("r", r, min=0),
-                "x0": lmfit.Parameter("x0", x0, min=0),
+                "x0": lmfit.Parameter("x0", x0, min=0.1),
                 "p": lmfit.Parameter("p", p, min=0),
-                "w": lmfit.Parameter("w", w, min=0),
+                "w": lmfit.Parameter("w", w, min=0.1),
```
Tom-Willemsen, line 756: These changes to `min` don't look right! were they just for testing?

src/ibex_bluesky_core/utils.py
```diff
@@ -104,3 +109,63 @@ def calculate_polarisation(
     polarisation_value.variances = variance_return
 
     return polarisation_value
+
+
+def calculate_erf_stretch(
```
Tom-Willemsen, line 114: Do we need to expose this publically, or should it just be a private helper method within `_fitting.py`?

src/ibex_bluesky_core/fitting.py
Tom-Willemsen: Unless you were having an actual problem with this, set to zero, or float_min if you _must_ I guess. But a fit with width or `x0` < 0.1 is entirely plausible.

Issue #152, opened 2025-03-27
In period-per-point mode make `run_number` a metadata key


Tom-Willemsen:
Or maybe unconditionally, but under `initial_run_rumber` or something in that case

PR #277, opened 2025-10-29
Make run_number a metadata key in period-per-point mode & document/make it easier to add scan-specific metadata

Tom-Willemsen:
### Description of work

- Makes `run_number` a metadata key in our standard `scan()` plan, if in period-per-point mode and `save_run` is True.
- Document how to pass around extra md so that users can add any _other_ metadata that they happen to want easily.
- Allow passing extra `md` into most of our top-level plans (as per bluesky upstream recommendation)

### Ticket

Closes https://github.com/ISISComputingGroup/ibex_bluesky_core/issues/152

### Acceptance criteria

- [x] Pull request title is understandable for a user (e.g. scientist) reading the release notes. The PR title should be a short description of the change from a user perspective.
- [x] Pull request has appropriate labels for automatic release-notes generation
- [x]  `run_number` is a metadata key in our standard `scan()` plan, if in period-per-point mode and `save_run` is True. Otherwise it should not be present.

### Documentation

Added documentation in the getting-started guide about how to add further metadata if users are not using "our" scan plan wrapper.

### DIFF
--- a/src/ibex_bluesky_core/plans/__init__.py
+++ b/src/ibex_bluesky_core/plans/__init__.py
@@ -53,7 +53,7 @@ def scan(  # noqa: PLR0913
     periods: bool = True,
     save_run: bool = False,
     rel: bool = False,
-    md: dict | None = None,
+    md: dict[Any, Any] | None = None,
 ) -> Generator[Msg, None, ISISCallbacks]:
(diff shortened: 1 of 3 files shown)


danielmaclaren (approved):
description in documentation, updated standard scans and test case passed

